PR #12, opened 2023-08-18
[WIP] Repl rework

emilypi:
Draft until it's done

- Rework modules into `Types.hs`
- Split out utils from datatypes 
- Split out commands into clear actions


emilypi:
Closing due to stale

PR #17, opened 2023-09-20
Get the typed core to compile

0xd34df00d:
There's plenty of work to be done still, but at least it compiles now, and I'd like to get it merged to avoid merge conflicts in the future.

There are also some a bit questionable design decisions I had to make, so we might want to discuss that later.


jmcardon:
This PR is great and I will cherry pick some changes. It's a bit too much work to revive with all the changes to the core compiler.

PR #39, opened 2023-12-11
Move parser to use more fv-accurate parsing

jmcardon:
This PR enforces more strictness on where `property` and `invariant` can show up vs master.

TODO: we have a regression on `meta.repl`... Should we then PR this?


jmcardon:
Closing this PR. We are using another approach via `PropertyExpr`

PR #40, opened 2023-12-11
Set limits on currently lexed/parsed params

jmcardon:
Note: `Analysis.Term` wasn't being used at all. It seems I added it months ago thinking I wanted another separate analysis IR. 

I think this is a good idea, but it's currently unused, so I am deleting it.

This PR:
- [x] Refactors `with-capability` to use our current `CapToken` machinery. It's easier that way, maybe we can get rid of `CapInvokeC` later.
- [x] sets a bunch of limits for identifiers, function args, etc.


jmcardon:
Outdated, will reopen

PR #73, opened 2024-01-17
return binop cast behavior on +-/*^log

jmcardon:
This PR removes our decision to remove the casting behavior on binary operators for decimals

* [x] Test coverage for the proposed change (See: ops.repl)


emilypi (commented):
It's kind of wonky that we upcast `*` to be decimal-typed in every case except integer args. Do we want to emit a deprecation warning for the March service date deprecating this behavior?

emilypi (approved):

jmcardon:
> It's kind of wonky that we upcast * to be decimal-typed in every case except integer args. Do we want to emit a deprecation warning for the March service date deprecating this behavior?

Indeed. I will do so in a followup.

pact-core/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -148,20 +153,49 @@ rawAdd info b cont handler _env = \case
...
+  [VLiteral (LDecimal i), VLiteral (LDecimal i')] ->
+    decimalSub i i'
+  [VLiteral (LInteger i), VLiteral (LDecimal i')] ->
+    decimalSub (Decimal 0 i) i'
+  [VLiteral (LDecimal i), VLiteral (LInteger i')] ->
+    decimalSub i (Decimal 0 i')
   args -> argsError info b args
+  where
+  decimalSub i i' = do
+    chargeGasArgs info (GIntegerOpCost PrimOpAdd (decimalMantissa i) (decimalMantissa i'))
```
0xd34df00d, line 176: Should this be `PrimOpAdd`? Addition and subtraction are indeed very similar, so if that's intended, I'd probably rename `PrimOpAdd` → `PrimOpAddSub`.

0xd34df00d (approved):

pact-core/Pact/Core/IR/Eval/CoreBuiltin.hs
jmcardon: Great suggestion!! Will do

PR #75, opened 2024-01-22
Explicitly enforce `describe-table` top-levelness

0xd34df00d:
The existing `repl` tests checking that `describe-table` only happens at top-level pass by semi-coincidence: the table isn't considered a pact value, so that check errors out. This PR adds an explicit check that `describe-table` is only called at top-level, instead of relying on what constitutes a pact value.

PR checklist:

* [x] Test coverage for the proposed changes (_in a separate failure tests branch_).
* [x] Any changes that could be relevant to users [have been recorded in the changelog](https://github.com/kadena-io/pact-core/blob/master/CHANGELOG.md)

* [ ] Benchmark regressions


rsoeldner (approved):

emilypi (commented):
Do we have a regression for this?

0xd34df00d:
> Do we have a regression for this?

@emilypi it shouldn't affect any existing code since it's already been an error (although for a different reason). This just makes this error more precise.

PR #82, opened 2024-01-23
fix repl completion of repl natives

### LINKED ISSUE #78: The autocompletion doesn't work with REPL natives
## Issue description

The autocompletion works well with "real Pact natives" but doesn't work with REPL only commands.
Example:

```pact
pact> (expec
```
I press TAB but nothing happens... I would expect the autocompletion to propose me `(expect  )`, `(expect-failure  )`

rsoeldner:
closes #78 

This PR also fixes same missing AST traversals resulting in bad position matching.

PR checklist:

* [ ] Benchmark regressions

---
  - https://app.asana.com/0/0/1206473309875668
  - https://app.asana.com/0/0/1206417680988244

### DIFF
--- a/pact/Pact/Core/Builtin.hs
+++ b/pact/Pact/Core/Builtin.hs
@@ -7,18 +7,17 @@
 
 module Pact.Core.Builtin
  ( CoreBuiltin(..)
- , coreBuiltinToText
  , coreBuiltinMap
  , coreBuiltinNames
  , ReplBuiltin(..)
- , replcoreBuiltinNames
- , replcoreBuiltinMap
- , replBuiltinToText
+ , replCoreBuiltinNames
+ , replCoreBuiltinMap
+ , coreBuiltinToUserText
+ , replCoreBuiltinToUserText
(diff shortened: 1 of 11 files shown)


pact/Pact/Core/Builtin.hs
```diff
@@ -288,7 +288,7 @@ coreBuiltinToText = \case
   CoreDistinct -> "distinct"
   CoreFormat -> "format"
   CoreEnumerate -> "enumerate"
-  CoreEnumerateStepN -> "enumerate-step"
+  CoreEnumerateStepN -> ""
```
rsoeldner, line 291: @jmcardon I feel like going through all builtins (+ overloads) should be a separate ticket. wdyt?

jmcardon: This changes hashes! Do not do this. We need to separate this out

jmcardon (changes requested):

jmcardon (approved):

PR #89, opened 2024-01-30
Fix take/drop and Obj pretty printing

### LINKED ISSUE #86: Strange behavior of negative `(drop)`
I experiment some incorrect behavior of the `(drop)` function.

```pact
pact>(drop 0 "ABC")
"ABC"
; This is OK

pact>(drop -1 "ABC")
"AB"
; This is OK

pact>(drop -2 "ABC")
"A"
; This is OK

pact>(drop -3 "ABC")
""
; This is OK

pact>(drop -4 "ABC")
"A"
; !! NOK !!

pact>(drop -5 "ABC")
"AB"
; !! NOK !!
```
It looks like that when a negative drop is larger than the string, there is a kind of "roll over". I would expect an empty string to be returned (like in Pact legacy)

Note that the behavior is correct with positive drops.

### LINKED ISSUE #87: Objects are not stringified to a valid JSON
In Pact legacy, objects were stringified as valid JSON:

**Example (Pact legcacy):**
```pact
pact> (format "{}" [{'a:3}])
"{"a": 3}"
pact> (print {'a:3})
{"a": 3}
```

But in Pact core, it's not longer the case.

**Example (Pact core):**
```pact
pact>(format "{}" [{'a:3}])
"{a:3}"
pact>(print {'a:3})
{a:3}
()
```
Note this missing "quotes" for the field name.

jmcardon:
See: https://app.asana.com/0/1204069832833999/1206459570889561

Closes #86 
Closes #87 

This PR fixes the following:

Take/drop on strings behaving incorrectly due to incorrect bounds checks. The following is output from the repl after the fixes in this PR.

```
pact>(drop -4 "ABC")
""
pact>(drop -5 "ABC")
""
```

Object pretty printing

```
pact>(format "{}" [{'a:3}])
"{"a":3}"
pact>(print {'a:3})
{"a":3}
()
```

PR checklist:

### DIFF
--- a/pact/Pact/Core/PactValue.hs
+++ b/pact/Pact/Core/PactValue.hs
@@ -91,10 +91,10 @@ instance Pretty PactValue where
     PObject o ->
       braces $ hsep $ punctuate comma (objPair <$> M.toList o)
       where
-      objPair (f, t) = pretty f <> ":" <> pretty t
+      objPair (f, t) = dquotes (pretty f) <> ":" <> pretty t
     PModRef md -> pretty md
     PCapToken (CapToken fqn args) ->
-      parens (pretty fqn) <> if null args then mempty else hsep (pretty <$> args)
+      parens (pretty (fqnToQualName fqn)) <> if null args then mempty else hsep (pretty <$> args)
     PTime t -> pretty (PactTime.formatTime "%Y-%m-%d %H:%M:%S%Q %Z" t)
 
 synthesizePvType :: PactValue -> Type
--- a/pact-tests/pact-tests/ops.repl
+++ b/pact-tests/pact-tests/ops.repl
@@ -508,6 +508,7 @@
 (expect "shift -255 -1" -128 (shift -255 -1))
 
 "===== drop"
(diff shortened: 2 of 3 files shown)


pact/Pact/Core/PactValue.hs
```diff
@@ -91,10 +91,10 @@ instance Pretty PactValue where
     PObject o ->
       braces $ hsep $ punctuate comma (objPair <$> M.toList o)
       where
-      objPair (f, t) = pretty f <> ":" <> pretty t
+      objPair (f, t) = dquotes (pretty f) <> ":" <> pretty t
     PModRef md -> pretty md
     PCapToken (CapToken fqn args) ->
-      parens (pretty fqn) <> if null args then mempty else hsep (pretty <$> args)
+      parens (pretty (fqnToQualName fqn)) <> if null args then mempty else hsep (pretty <$> args)
```
rsoeldner, line 97: :+1:

rsoeldner (approved):

PR #90, opened 2024-01-30
Fix doc pretty printing

### LINKED ISSUE #85: Issues with top-level constants
I've discovered that Pact core allows the use of top-level constants and functions.
A great improvement. Hope this will be available for transactions as well.

Not sure, top-level consts are supposed to be functional at this point
but it seems there are still some issues:

**Issue 1:**
When directly used, constants are recognized as functions, and they are pretty-printed instead of being directly evaluated.
```pact
pact>(defconst MY-CONST:integer 3)
Loaded defconst MY-CONST

pact>MY-CONST
function #repl.MY-CONST:
``` 

**Issue 2:**
When used indirectly, they still don't work:

```pact
pact>(defconst MY-CONST:integer 3)
Loaded defconst MY-CONST

pact>(+ 1 MY-CONST)
(interactive):1:5: Fatal execution error, invariant violated: Defconst not fully evaluated
 1 | (+ 1 MY-CONST)
   |      ^^^^^^^^
```

jmcardon:
In this pr, the following in a repl script

```

(module m g
  (defcap g ()
    @doc "bloop"
    true)

  (defun dfun-example:integer (a:integer)
    @doc "these are the defun docs"
    (+ a 1)
  )

  (defcap dcap-example:integer (a:integer)
    @doc "these are the defcap docs"
    (+ a 1)
  )

  (defpact dpact-example:integer (a:integer)
    @doc "these are the defpact docs"
    (step 1)
  )

  (defconst dconst-example:integer 1
    @doc "this is the defconst example"
    )

  (defschema schema-example
    @doc "this is the schema doc"
    a:integer)

  (deftable tbl-example:{schema-example}
    @doc "this is the table docs"
    )
  )

dconst-example
dfun-example
dcap-example
dpact-example
tbl-example
```

Produces
```
Loaded module m, hash LmLumnFm1auzfNPEMEbK8Q0FufFJ2sMqTGsuq2wfuqw
(defconst dconst-example:integer 1)
Docs:
this is the defconst example
(defun dfun-example:integer (a:integer))
Docs:
these are the defun docs
(defcap dcap-example:integer (a:integer))
Docs:
these are the defcap docs
(defpact dpact-example:integer (a:integer))
Docs:
these are the defpact docs
(deftable tbl-example:m.schema-example)
Docs:
this is the table docs
```

Closes #85 

Note: schema docs are currently not being spit out because we fail at compile-time for them, as they are invalid in variable positions. We should maybe turn this into a runtime failure.

PR checklist:

* [ ] Any changes that could be relevant to users [have been recorded in the changelog](https://github.com/kadena-io/pact

### DIFF
--- a/pact/Pact/Core/IR/ModuleHashing.hs
+++ b/pact/Pact/Core/IR/ModuleHashing.hs
@@ -110,7 +110,7 @@ encodeModule (Module mname mgov defs mblessed imports mimps _mh _mi) = parens $
   <> (if null mblessed then mempty else hsep (encodeBless <$> S.toList mblessed))
   <> (if null imports then mempty else hsep (encodeImport <$> imports) )
   <> (if null mimps then mempty else hsep (encodeModuleName <$> mimps))
-  <> hsep (encodeDef <$> defs)
+  <> hsep (encodeDef mname <$> defs)
   where
   encodeGov :: Governance Name -> Builder
   encodeGov (KeyGov (KeySetName name mNs)) = encodeMNamespace mNs <> encodeText name
@@ -126,7 +126,7 @@ encodeInterface (Interface ifn idefs imports _h _i) = parens $
   "interface"
   <+> encodeModuleName ifn
   <> (if null imports then mempty else hsep (encodeImport <$> imports) )
-  <> hsep (encodeIfDef <$> idefs)
+  <> hsep (encodeIfDef ifn <$> idefs)
 
 encodeText :: T.Text -> Builder
 encodeText = T.encodeUtf8Builder
@@ -223,11 +223,8 @@ encodeApp operator operands =
   parens $ hsep (operator:operands)
 
 encodeSchema :: Schema -> Builder
-encodeSchema (Schema sc) =
-  hsep (encodePair <$> M.toList sc)
-  where
-  encodePair (Field f, ty) =
-    T.encodeUtf8Builder f <> ":" <> encodeType ty
+encodeSchema (Schema n _sc) =
+  encodeQualName n
 
 encodePrim :: PrimType -> Builder
 encodePrim = \case
@@ -330,12 +327,12 @@ encodeTerm = \case
 encodeTyAnn :: Maybe Type -> Builder
 encodeTyAnn = maybe mempty ((":" <>) . encodeType)
 
-encodeDef :: IsBuiltin b => Def Name Type b i -> Builder
-encodeDef = \case
+encodeDef :: IsBuiltin b => ModuleName -> Def Name Type b i -> Builder
+encodeDef mn = \case
   Dfun d -> encodeDefun d
   DConst d -> encodeDefConst d
   DCap d -> encodeDefCap d
-  DSchema s -> encodeDefSchema s
+  DSchema s -> encodeDefSchema mn s
   DTable d -> encodeDefTable d
   DPact d -> encodeDefPact d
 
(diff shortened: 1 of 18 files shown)


pact/Pact/Core/IR/ModuleHashing.hs
```diff
@@ -223,11 +223,8 @@ encodeApp operator operands =
   parens $ hsep (operator:operands)
 
 encodeSchema :: Schema -> Builder
-encodeSchema (Schema sc) =
-  hsep (encodePair <$> M.toList sc)
-  where
-  encodePair (Field f, ty) =
-    T.encodeUtf8Builder f <> ":" <> encodeType ty
+encodeSchema (Schema n _sc) =
+  encodeQualName n
```
jmcardon, line 227: note: this change is quite significant. It means our stable encoding of modules do not blow up nested schemas in types in size, so the output serialization stays small. Schemas are basically serialized as free variable references which is what we want.

CryptoPascal31:
@jmcardon 
IMHO : ` dconst-example`  should produce the value of the constant (like in legacy Pact and LISP ). Not 100% sure about that.. Because harmonizing with functions, tables, caps... and always print doc make senses too. 

Question: What will be the behavior of the same script in a real transaction, not in REPL... Produces the doc too ?

jmcardon:
@CryptoPascal31 I personally like what I did just now. It spits out the full doc for a defconst, which is fine, but they still work as regular terms. For example:

```
(defconst dconst-example:integer 1)
Docs:
this is the defconst example
2
```

> Question: What will be the behavior of the same script in a real transaction, not in REPL... Produces the doc too ?
That exact script would actually fail. Spitting out docs on-chain is disallowed now in pact. 

The doc script handling is done [here](https://github.com/kadena-io/pact-5/pull/90/files#diff-7f5118a255d3c15b6399efff96e9793ff52575657d8362a6abc8be62b2bd5d39R140), with the snippet: 
```
        case ds of
          TLTerm (Var (Name n (NTopLevel mn mh)) varI) -> do
            let fqn = FullyQualifiedName mn n mh
            lookupFqName fqn >>= \case
              Just d -> do
                let qn = QualifiedName n mn
                docs <- uses replUserDocs (M.lookup qn)
                displayValue (RUserDoc d docs)
              Nothing ->
                failInvariant varI "repl invariant violated: resolved to a top level free variable without a binder"
          _ -> do
            v <- evalTopLevel replEnv ds deps
            displayValue (RCompileValue v)
```

This is exclusively for the repl. In a transaction, it would simply only call the `evalTopLevel` branch. Since closures are not allowed as return values @ the top level, an expression like a lone `dfun-example` at the top level will cause it to fail.

If it would be useful to have a repl "mode" that would behave exactly as a tx would, I can add this to aid users, please file an issue if so.

jmcardon:
Note: defconst usage has also been fixed in the repl as well.

pact/Pact/Core/IR/Term.hs
```diff
@@ -265,23 +266,22 @@ findDefInModule :: Text -> Module name ty b i -> Maybe (Def name ty b i)
 findDefInModule defnName targetModule =
   find ((==) defnName . defName) (_mDefs targetModule)
 
-defKind :: Def name Type b i -> DefKind
-defKind = \case
+defKind :: ModuleName -> Def name Type b i -> DefKind
+defKind mn = \case
```
jmcardon, line 270: this is a little cheeky internally :). That said: `defKind` is always called when the module name is in-scope, so this is fine.

rsoeldner (approved):

PR #100, opened 2024-02-03
Derive `Data` and avoid hand-writing `Plated Term` instance

0xd34df00d:
Just a proposal to avoid writing `Plated` for `Term` (and potentially other types) manually, which is a tad tedious. Feel free to frown upon!

---

PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/IR/Term.hs
+++ b/pact/Pact/Core/IR/Term.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DeriveDataTypeable #-}
 {-# LANGUAGE DeriveFunctor #-}
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE TemplateHaskell #-}
@@ -22,6 +23,7 @@
 module Pact.Core.IR.Term where
 
 import Control.Lens
+import Data.Data(Data)
 import Data.Foldable(fold, find)
 import Data.Text(Text)
 import Data.List.NonEmpty (NonEmpty)
@@ -76,7 +78,7 @@ data Term name ty builtin info
   -- ^ an object literal
   | CapabilityForm (CapForm name (Term name ty builtin info)) info
   -- ^ Capability Natives
-  deriving (Show, Functor, Eq, Generic)
+  deriving (Show, Functor, Eq, Generic, Data)
 
 data ConstVal term
   = TermConst term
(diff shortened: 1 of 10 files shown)


pact/Pact/Core/IR/Term.hs
```diff
@@ -542,26 +544,7 @@ traverseDefTerm f = \case
...
-    ListLit m i -> ListLit <$> traverse f m <*> pure i
-    Nullary term i ->
-      Nullary <$> f term <*> pure i
-    CapabilityForm cf i ->
-      CapabilityForm <$> traverse f cf <*> pure i
-    Try e1 e2 i ->
-      Try <$> f e1 <*> f e2 <*> pure i
-    ObjectLit o i ->
-      ObjectLit <$> (traverse._2) f o <*> pure i
+instance (Data ty, Data name, Data builtin, Data info) => Plated (Term name ty builtin info)
```
0xd34df00d, line 547: Could've also done
```
deriving (..., Plated)
```
at `Term` definition site, but `DeriveAnyClass` isn't generally used in the codebase, so dunno if you'd be happy with that.

pact/Pact/Core/Literal.hs
```diff
@@ -19,13 +24,15 @@ import GHC.Generics
 
 import Pact.Core.Pretty
 
+deriving instance Data Decimal
```
0xd34df00d, line 27: This orphan will go away if/when a PR I opened earlier to `data-decimal` will be merged.

I'm also happy to move this orphan to a separate orphans module, since we already have another orphan elsewhere, so we might as well isolate that.

rsoeldner: A `Todo` would be great, so that it does not get lost.

rsoeldner (approved):
:+1: I feel hand-writing the `Plated` instance is error-prone.

jmcardon (changes requested):
I'm not sure how I feel. 

Yeah, it's error prone, but there's only one instance where this can happen, and the only reason we're even using a `Plated` instance is for one traversal in `ModuleHashing.hs`. We could, ostensibly, just do without `Plated` entirely, by simply writing the traversal by hand.

We're deriving something like 10+ data instances that replace only 15 lines of code, as well as changing constraints on a few callsites.

jwiegley:
My sense then would be to write the traversal.

0xd34df00d:
The other option to save on the boilerplate is to use TH to derive the required `transform` via [geniplate](https://hackage.haskell.org/package/geniplate-mirror-0.7.9/docs/Data-Generics-Geniplate.html#v:genTransformBi) which avoids having to derive `Data` and orphan instances and what not. On the negative side, it adds an extra dependency on what I'd call a less popular library, though.

> as well as changing constraints on a few callsites like requiring Data for just a Show instance

Ah, if you're talking about `IsBuiltin`, I'm just adding the `Data` constraint to the context, to avoid spelling it out everywhere `IsBuiltin` is mentioned. It's not needed for `Show` unless I misunderstood you.

jmcardon:
Closing for now, but the idea may be useful in the future

PR #103, opened 2024-02-07
docs(core): extend the grammar definition

lukeribchester:
This pull request extends the Pact grammar definition documented in `Semantics.md`.

**Draft:** The current definition is approximately complete based on the current Pact 5 parser implementation, but contains a few minor issues which are visible when testing against the [Pact test files](https://github.com/kadena-io/pact-5/tree/master/pact-tests/pact-tests). These issues will be documented, discussed, and fixed here before merging this pull request.

### To Do

- [ ] Define operators (numerical, logical, bitwise etc.).
- [ ] Define `use` keyword for imports.
- [ ] Add important regular expressions (comments, strings etc.).
- [ ] Fix if-statement definition.
- [ ] Fix lamba definition.

### Pull Request Checklist

### DIFF
--- a/pact/Pact/Core/Semantics.md
+++ b/pact/Pact/Core/Semantics.md
@@ -23,8 +23,8 @@ in [Extended Backus-Naur form (EBNF)](https://en.wikipedia.org/wiki/Extended_Bac
 ### Lexical Syntax
 
 ```
-lower          ::= 'a' | ... | 'z' (Ascii only, no unicode lower case)
-upper          ::= 'A' | ... | 'Z' (Ascii only, no unicode lower case)
+lower          ::= 'a' | ... | 'z' // Ascii only, no unicode lower case
+upper          ::= 'A' | ... | 'Z' // Ascii only, no unicode lower case
 special-symbol ::= '%' | '#' | '+' | '-' | '_' | '&' | '$' | '@'
                      | '<' | '>' | '=' | '^' | '?' | '*' | '!'
                      | '|' | '/' | '~'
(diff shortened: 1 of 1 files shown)


jmcardon:
Hey @lukeribchester, thanks for the PR! Looks good so far, let me know what issues come up, because the current parser spec works against all the test files.

lukeribchester:
> Hey @lukeribchester, thanks for the PR! Looks good so far, let me know what issues come up, because the current parser spec works against all the test files.

Hey @jmcardon, thank you and apologies for my delayed reply. I've further refined the grammar definition based on my experiences working with it for the IntelliJ Platform plugin release, and I've separated most special characters from the identifier token definition into their respective definitions and categories. My parser now functions correctly against all test files except for [`fv-syntax-regression.repl`](https://github.com/kadena-io/pact-5/blob/master/pact-tests/pact-tests/fv-syntax-regression.repl) — where individual `%` and `@` symbols raise syntax errors. My understanding is that neither of these are operators in Pact nor Lisp. I could classify them as identifiers but that feels weird. What do you recommend?

lukeribchester:
This is currently in use in version `0.2.0` of the [Pact plugin](https://plugins.jetbrains.com/plugin/23581-pact) for the IntelliJ Platform if you wish to check it out — it's the first release with native language support.

jmcardon:
> > Hey @lukeribchester, thanks for the PR! Looks good so far, let me know what issues come up, because the current parser spec works against all the test files.
> 
> Hey @jmcardon, thank you and apologies for my delayed reply. I've further refined the grammar definition based on my experiences working with it for the IntelliJ Platform plugin release, and I've separated most special characters from the identifier token definition into their respective definitions and categories. My parser now functions correctly against all test files except for [`fv-syntax-regression.repl`](https://github.com/kadena-io/pact-5/blob/master/pact-tests/pact-tests/fv-syntax-regression.repl) — where individual `%` and `@` symbols raise syntax errors. My understanding is that neither of these are operators in Pact nor Lisp. I could classify them as identifiers but that feels weird. What do you recommend?


Hi I'm sorry for the delayed reply!

Pact-5 is just parsing this atm, and we amended the parse tree to work with this regression.

PR #107, opened 2024-02-14
Step debugger

0xd34df00d:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/Info.hs
+++ b/pact/Pact/Core/Info.hs
@@ -1,6 +1,7 @@
 module Pact.Core.Info
  ( SpanInfo(..)
  , combineSpan
+ , isWithin
  ) where
 
 import Data.Default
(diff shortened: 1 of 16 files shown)


pact/Pact/Core/Info.hs
```diff
@@ -22,3 +23,15 @@ instance Default SpanInfo where
 combineSpan :: SpanInfo -> SpanInfo -> SpanInfo
 combineSpan (SpanInfo l1 c1 _ _) (SpanInfo _ _ l2 c2) =
   SpanInfo l1 c1 l2 c2
+
```
rsoeldner, line 26: We have this as part of the LSP: https://github.com/kadena-io/pact-5/blob/b3f2c86afde946086fb123f100d2352230ff0e64/pact-lsp/Pact/Core/LanguageServer/Utils.hs#L104

pact/Pact/Core/Info.hs
```diff
@@ -1,37 +1,74 @@
...
-  -- similarly for the last line of the span:
-  | l == el = c <= ec
-  -- otherwise the position could only be between the starting and ending lines
-  -- or outside the span, and columns need not be checked:
-  | otherwise = l > sl && l < el
+combineSpan (SpanInfo s _) (SpanInfo _ e) = SpanInfo s e
+
+inside :: Position -> SpanInfo -> Bool
+inside p (SpanInfo s e) = s <= p && p < e
+
```
rsoeldner, line 44: I like this change:+1

jmcardon:
Closing this branch: stale. Step debugger work will resume in the future

PR #117, opened 2024-03-15
Add some strictness annotations

0xd34df00d:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/gasmodel/Pact/Core/GasModel/InterpreterGas.hs
+++ b/gasmodel/Pact/Core/GasModel/InterpreterGas.hs
@@ -1300,7 +1300,7 @@ gasContType pdb = \case
     gasLetC pdb
   CTSeqC ->
     gasSeqC pdb
-  CTListC -> undefined
+  CTListC -> C.bgroup "list" []
   -- Conditionals
   CTAndC ->
     C.bgroup "AndC Cases" $ (gasAndC pdb) <$> [minBound .. maxBound]
@@ -1312,7 +1312,7 @@ gasContType pdb = \case
       | b <- [False, True] ]
   CTEnforceC ->
     C.bgroup "EnforceC Cases" $ (gasEnforceC pdb) <$> [minBound .. maxBound]
-  CTEnforceOneC -> undefined
+  CTEnforceOneC -> C.bgroup "enforce one" []
   CTFilterC ->
     C.bgroup "FilterC Cases" $
       [gasFilterCEmptyNElems pdb b 10
--- a/gasmodel/Pact/Core/GasModel/InterpreterGas.hs
+++ b/gasmodel/Pact/Core/GasModel/InterpreterGas.hs
@@ -23,6 +23,7 @@ import qualified Database.SQLite3 as SQL
 
 import Pact.Core.Builtin
 import Pact.Core.Environment
+import Pact.Core.Errors
 import Pact.Core.Names
 import Pact.Core.Gas
 import Pact.Core.Literal
@@ -39,6 +40,13 @@ import qualified Pact.Core.IR.Eval.CEK as Eval
 
 import Pact.Core.GasModel.Utils
 
+
+runEvalDropState
+  :: EvalEnv b i
+  -> EvalState b i
+  -> EvalM b i a
+  -> IO (Either (PactError i) a)
+runEvalDropState ee es = fmap fst . runEvalM ee es
 
 benchmarks :: C.Benchmark
 benchmarks = C.envWithCleanup mkPactDb cleanupPactDb $ \ ~(pdb, _db) -> do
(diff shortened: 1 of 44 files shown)


gasmodel/Pact/Core/GasModel/InterpreterGas.hs
```diff
@@ -1300,7 +1300,7 @@ gasContType pdb = \case
     gasLetC pdb
   CTSeqC ->
     gasSeqC pdb
-  CTListC -> undefined
+  CTListC -> C.bgroup "list" []
```
0xd34df00d, line 1303: We'll add more benchmarks later, but for now `undefined` just breaks running all the benchmarks or matching on a pattern.

jwiegley:
Where are the performance numbers to show that this is beneficial in terms of either memory or CPU?

jwiegley (changes requested):
Need performance numbers to show that this is beneficial in terms of either memory or CPU.

emilypi:
Generally I'm wary of tossing `{-# StrictData #-}` in places as a performance improvement because it's mostly a parser-phase `!` insertion on records are sums. I would _like_ any benchmarking to include some analysis of predictability of the priors instead of just the blunt instrument of "goes faster/looks nicer". We need to know our cost centers!

0xd34df00d:
Sure, here are the benchmark outputs:
* for the current `master`: [gist](https://gist.github.com/0xd34df00d/aa5252ad9e05967c324c8bb1d265328f)
* for the proposed changes: [gist](https://gist.github.com/0xd34df00d/ccf06baf74d9e1df6529aa56e314fb41)

Here are also the relative differences (with the denominator being the average of the old and new number), 118 out of 122 benchmarks in favor of the proposed changes: [gist](https://gist.github.com/0xd34df00d/b544829b5b67ea8a343f92890160fcf3) (and the not-in-favor being less -4.6%, -2.3%, -0.9% and -2.4% differences). So it's mostly two-digit relative performance improvements, up to 80% on some tests.

It's also worth noting that these benchmarks might be hard to interpret unambiguously. `nfAppIO` forces each result of all the `runEvalM`s to normal form, but that result contains the interpreter state, which, in particular, includes the `esLoaded` field. This state is not really _the_ output of interpretation, but IMO it's worth including the time to evaluate it in the benchmarks (as it indeed currently happens), otherwise it's harder to reason about performance of the compositions of the functions (since `esLoaded` is used during evaluation, and some parts of `esLoaded` might be lazily delayed to be evaluated later when a subsequent function uses them, so _not_ counting them might be sort of cheating).

What's ambiguous about all that? Well, `esLoaded` includes non-trivial (Haskell) terms, so even with the strictness annotations it takes a non-trivial amount of time just to ensure that it's in normal form. For example, replacing
```haskell
C.bench "(+ 1 2)" $ C.nfAppIO (runEvalM ee' es' . Eval.evalNormalForm env') term'
```
with
```haskell
C.bench "(+ 1 2)" $ C.nfAppIO (fmap (second force) . runEvalM ee' es' . Eval.evalNormalForm env') term'
```
(which gives the same result, but with an extra nf'ing pass over the interpreter state) increases the run time from 290-300 ns to about 440-460 ns (all with strictness annotations), which might be a good ballpark estimate of the overhead of just traversing the interpreter state to ensure it's in NF (although, for this particular test!). So, for this particular case it's arguably more correct to interpret all these numbers as traversing the interpreter state taking about 150 ns, and thus pure interpretation time for the `master` version is 380 - 150 = 230 ns and for the "strict" version it's 290 - 150 = 140 ns, giving about 48% perf improvement (instead of 26%, if criterion report numbers are taken as is).

State traversal overhead is different for different tests for a bunch of reasons, so the best approach to account for that would be to have a version of each benchmark with and without an additional `force`, but I'm not sure it's worth investing time right now.

@emilypi
> Generally I'm wary of tossing {-# StrictData #-} in places as a performance improvement because it's mostly a parser-phase ! insertion on records are sums.

That's in fact exactly what I want here, since I caught myself just inserting lots of `!`s in record fields that are transitively involved in `Loaded`/`EvalState`.

> We need to know our cost centers!

This particular case appears to have optimizations being quite sensitive to profiling builds. The difference between proposed changes and `master` are much less pronounced (and sometimes around 0%) when compiled with profiling (and all the usual auto-SCC jazz), even if profiling is disabled at runtime (that is, no `-p` and friends). I have recorded cost centers for some of our tests, but they aren't very telling (the test function itself typically being the biggest cost center), and the aforementioned change of behavior makes me consider them less reliable.

jwiegley (approved):
This is the kind of reasoning I like to see. If it's a measurable improvement, I have no issue with the changes presented. I will leave the decision up to @emilypi and @jmcardon from here.

jmcardon (approved):

PR #118, opened 2024-03-20
New stuff for the chainweb integration

ak3n:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/Errors.hs
+++ b/pact/Pact/Core/Errors.hs
@@ -345,6 +345,7 @@ data EvalError
   | EnforcePactVersionParseFailure Text
   | RuntimeRecursionDetected QualifiedName
   | SPVVerificationFailure Text
+  | ContinuationError Text
   deriving (Show, Generic)
 
 instance NFData EvalError
@@ -482,7 +483,7 @@ instance Pretty EvalError where
     e@MismatchingKeysetNamespace{} -> pretty (show e)
     e@RuntimeRecursionDetected{} -> pretty (show e)
     e@SPVVerificationFailure{} -> pretty (show e)
-
+    e@ContinuationError{} -> pretty (show e)
 
 
 instance Exception EvalError
--- a/pact/Pact/Core/SPV.hs
+++ b/pact/Pact/Core/SPV.hs
@@ -1,12 +1,19 @@
 module Pact.Core.SPV
- ( ContProof(..)
- , SPVSupport(..)
- , noSPVSupport)where
+  ( -- * Types
+    ContProof(..)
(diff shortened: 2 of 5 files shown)


ak3n:
@jmcardon i think it would nice to merge this PR

pact/Pact/Core/SPV.hs
```diff
@@ -1,22 +1,37 @@
...
 import Pact.Core.PactValue
 
 newtype ContProof = ContProof { _contProof :: ByteString }
   deriving (Eq, Ord, Show, Generic)
 
+instance Wrapped ContProof
 
 instance NFData ContProof
 
+instance J.Encode ContProof where
```
rsoeldner, line 28: Why do we need an `Encode` instance for `ContProof` ?

pact/Pact/Core/SPV.hs
```diff
@@ -1,22 +1,37 @@
...
 import Control.DeepSeq
+import Control.Lens (Wrapped)
 
 import Pact.Core.DefPacts.Types
 import Pact.Core.PactValue
 
 newtype ContProof = ContProof { _contProof :: ByteString }
   deriving (Eq, Ord, Show, Generic)
 
+instance Wrapped ContProof
```
rsoeldner, line 24: What is the reason for it?

pact/Pact/Core/Errors.hs
```diff
@@ -345,6 +345,7 @@ data EvalError
   | EnforcePactVersionParseFailure Text
   | RuntimeRecursionDetected QualifiedName
   | SPVVerificationFailure Text
+  | ContinuationError Text
```
rsoeldner, line 348: Should probably be renamed to `SPVContinuationError`?

pact/Pact/Core/SPV.hs
jmcardon: Most likely to encode it in outputs

jmcardon (approved):

rsoeldner (approved):

pact/Pact/Core/Errors.hs
emilypi: Yeah. The defpact cont errors are above, and this is presumably a proof function error from `yield/resume`.

pact/Pact/Core/SPV.hs
emilypi: No need.

pact/Pact/Core/SPV.hs
emilypi: It wouldn't go into outputs, but it would tx data that we serialize.

PR #121, opened 2024-04-22
Publish binary as development-latest

rsoeldner:
This PR adds releasing binary artefacts once they pass CI as `development-latest`. The general idea is to ship new versions of the LSP server via the VSCode plugin.

PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -44,7 +44,7 @@ jobs:
 
         echo Build the bundle
         nix build .#pact-binary-bundle --log-lines 500 --show-trace --out-link pact-binary-bundle
-        tar -zcvf pact-binary-bundle-.${{ matrix.os }}.tar.gz ./pact-binary-bundle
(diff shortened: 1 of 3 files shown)


jmcardon (approved):

jmcardon:
Please squash all the wip commits together

.github/workflows/nix.yml
```diff
@@ -42,5 +42,24 @@ jobs:
         echo Building the project and its devShell
         nix build .#check --log-lines 500 --show-trace
 
+        echo Build the bundle
+        nix build .#pact-binary-bundle --log-lines 500 --show-trace --out-link pact-binary-bundle
+        tar -zcvf pact-binary-bundle.${{ matrix.os }}.tar.gz $(readlink pact-binary-bundle)
```
emilypi, line 47: In retrospect, we should probably decide on what this should be named and come up with a convention for it. `pact-binary-bundle.*.tar.gz` isn't very friendly or informative

PR #123, opened 2024-04-24
add identity builtin to support LSP

rsoeldner:
In this PR we add a dedicated builtin for the `identity` function. So-far, it was desugared directly into a lambda. This prevented us from displaying docs from the LSP.

PR checklist:

* [ ] Benchmark regressions

---
  - https://app.asana.com/0/0/1207151514407085

### DIFF
--- a/pact-tests/pact-tests/identity.repl
+++ b/pact-tests/pact-tests/identity.repl
@@ -1,0 +1,19 @@
+
+;; tests on literals
+(expect "identity on string literal" (identity "abc") "abc")
+(expect "identity on integer literal" (identity 123) 123)
+(expect "identity on decimal literal" (identity 12.3) 12.3)
+(expect "identity on bool literal: true" (identity true) true)
+(expect "identity on bool literal: false" (identity false) false)
+
+(expect-failure "identity on string literals can fail" (enforce (= (identity "a") "b")))
+
+;; tests on lists
+(expect "identity on list of literal" (identity ["a" "b" "c"]) ["a" "b" "c"])
+(expect "identity on nested list of literal" (identity [["a"]]) [["a"]])
+
+;; tests on object
+(expect "identity on object" (identity {"a": 1}) {"a": 1})
+
+;; tests on time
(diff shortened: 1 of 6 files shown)


imalsogreg:
This works well in normal cases:

pact-5:
```
(map (compose (+ 1) (identity)) [1,2])
[2, 3]
```

However the behavior is slightly different between legacy and pact-5. I'm not sure how important this matters:

legacy-pact:
```
(map identity [1,2])
<interactive>:0:0:Error: Invalid arguments, received
```

pact-5:
```
pact>(map identity [1,2])
[1, 2]
```

pact-5 is more permissive than legacy pact about the arguments to `map`.

This is probably an issue with `map` more than an issue with `identity`. And `pact-5:master` has the same behavior as `pact-5:rsoeldner/add-identity-builtin`, so this PR looks good to me.

imalsogreg (approved):
Manually tested with several happy-path and unhappy-path arguments. It behaves as expected.

There is a discrepancy between pact-5 and the legacy interpreter with expressions like `(map identity [1,2])`, but this is true both before and after the change in this PR. (not this PR's fault).

rsoeldner:
> This works well in normal cases:
> 
> pact-5:
> 
> ```
> (map (compose (+ 1) (identity)) [1,2])
> [2, 3]
> ```
> 
> However the behavior is slightly different between legacy and pact-5. I'm not sure how important this matters:
> 
> legacy-pact:
> 
> ```
> (map identity [1,2])
> <interactive>:0:0:Error: Invalid arguments, received
> ```
> 
> pact-5:
> 
> ```
> pact>(map identity [1,2])
> [1, 2]
> ```
> 
> pact-5 is more permissive than legacy pact about the arguments to `map`.
> 
> This is probably an issue with `map` more than an issue with `identity`. And `pact-5:master` has the same behavior as `pact-5:rsoeldner/add-identity-builtin`, so this PR looks good to me.

Great assessment! Thank you.

emilypi (approved):

PR #124, opened 2024-05-03
More benchmarks for natives

0xd34df00d:
PR checklist:

* [x] Any changes that could be relevant to users [have been recorded in the changelog](https://github.com/kadena-io/pact-core/blob/master/CHANGELOG.md)

* [ ] Benchmark regressions


gasmodel/Pact/Core/GasModel/BuiltinsGas.hs
```diff
@@ -77,10 +97,24 @@ benchArithBinOp' growthFactor op pdb =
     [ runNativeBenchmark pdb title [text|($op $x $x.0)|] | (title, x) <- vals ]
   ]
   where
-  vals = take 3 $ enumExpText 1000 growthFactor
+  vals = take 3 $ enumExpText 1000 1000000
```
rsoeldner, line 100: Maybe we should use `NumericUnderscores` and `1_000_000`?

```suggestion
  vals = take 3 $ enumExpText 1000 1_000_000
```

gasmodel/Pact/Core/GasModel/Utils.hs
```diff
@@ -322,21 +324,80 @@ runNativeBenchmark
...
+stAddDef :: Text -> Def Name Type CoreBuiltin () -> StMod
+stAddDef name dfn = Endo $ (esLoaded.loToplevel %~ M.insert name (fqn, dkind))
+                         . (esLoaded.loAllLoaded %~ M.insert fqn dfn)
+  where
+  fqn = mkGasModelFqn name
+  dkind = case dfn of
+    Dfun{} -> DKDefun
+    DConst{} -> DKDefConst
+    DCap{} -> DKDefCap
+    DSchema{} -> error "stAddDef: todo support schemas"
```
rsoeldner, line 379: leftover todo

pact/Pact/Core/Gas.hs
```diff
@@ -138,8 +141,35 @@ data IntegerPrimOp
   | PrimOpMul
   | PrimOpDiv
   | PrimOpShift
+  | PrimOpPow
   deriving (Eq, Show, Enum, Ord, Generic, NFData)
 
+data StrOp
```
rsoeldner, line 147: :+1:

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1268,26 +1325,32 @@ coreDistinct info b cont handler _env = \case
...
     alternate _ _ = []
-  args -> argsError info b args
 
--- Todo: This _Really_ needs gas
--- moreover this is kinda hacky
--- BIG TODO: REMOVE PRETTY FROM SEMANTICS.
--- THIS CANNOT MAKE IT TO PROD
-renderPactValue :: PactValue -> T.Text
-renderPactValue = T.pack . show . Pretty.pretty
+    -- Todo: this is kinda hacky
```
rsoeldner, line 1348: What happened here ?

pact/Pact/Core/IR/Eval/Runtime/Utils.hs
```diff
@@ -97,7 +97,8 @@ lookupFqName fqn =
 getDefCap :: (MonadEval b i m) => i -> FullyQualifiedName -> m (EvalDefCap b i)
 getDefCap info fqn = lookupFqName fqn >>= \case
   Just (DCap d) -> pure d
-  _ -> failInvariant info "Expected DefCap"
+  Just _ -> failInvariant info "Expected DefCap"
+  _ -> failInvariant info "Expected DefCap; got nothing"
```
rsoeldner, line 101: :+1: we should go over all the `failInvariant` invocations and enhance the messages.

rsoeldner (commented):
Nice work!

gasmodel/Pact/Core/GasModel/BuiltinsGas.hs
0xd34df00d: Good idea! I'll make a pass over the code to see what other greater-than-1000 numbers are there.

pact/Pact/Core/IR/Eval/Runtime/Utils.hs
0xd34df00d: Yep, that's on my todo list as a separate issue. We'll need to have a proper ADT for various invariant failures instead of just strings as they are now.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: I moved the function to be a local `where`-bound definition in the `coreFormat` (since it's not used elsewhere and it's a good way to show/prove that its scope is limited), and I also removed part of the comment since it's addressed now.

gasmodel/Pact/Core/GasModel/Utils.hs
0xd34df00d: Not yet! `DKDefSchema` requires a `Schema` (thus differing from the rest of `DefKind` ctors), and while I think it could be constructed from the definition itself, I didn't want to spend much time on this right now, since I don't have defschemas in my benchmarks so far. But shall we ever encounter those, we'll get an error message telling what to do!

If you feel like it's still worth addressing right now instead of evaluating lazily, lmk and I'll take a look.

gasmodel/Pact/Core/GasModel/BuiltinsGas.hs
0xd34df00d: Went through all of it and changed all numbers bigger than 1000 to have `_` separators. Great suggestion, thanks!

pact/Pact/Core/IR/Eval/Runtime/Utils.hs
jmcardon: Indeed. I agree, we need a better ADT for invariant failures.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1984,7 +2065,7 @@ coreBuiltinRuntime = \case
   CoreDefineNamespace -> coreDefineNamespace
   CoreDescribeNamespace -> coreDescribeNamespace
   CoreZkPairingCheck -> zkPairingCheck
-  CoreZKScalarMult -> zkScalaMult
+  CoreZKScalarMult -> zkScalarMult
```
jmcardon, line 2068: I like that we are now no longer multiplying Scalas :)

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1656,8 +1735,10 @@ coreDescribeNamespace :: (CEKEval step b i m, MonadEval b i m) => NativeFunction
 coreDescribeNamespace info b cont handler _env = \case
   [VString n] -> do
     pdb <- viewEvalEnv eePactDb
+    chargeGasArgs info $ GRead $ sizeOf SizeOfV0 n
```
jmcardon, line 1738: HMM. This does not seem right. Why are we charging gas on the size of the name of the namespace?

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1656,8 +1735,10 @@ coreDescribeNamespace :: (CEKEval step b i m, MonadEval b i m) => NativeFunction
...
     liftDbFunction info (_pdbRead pdb DNamespaces (NamespaceName n)) >>= \case
-      Just (Namespace _ usrG laoG) -> do
+      Just existing@(Namespace _ usrG laoG) -> do
+        chargeGasArgs info $ GRead $ sizeOf SizeOfV0 existing
```
jmcardon, line 1741: This one is definitely the correct one 👍

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1620,11 +1697,13 @@ coreDefineNamespace info b cont handler env = \case
     let pdb = view cePactDb env
     let nsn = NamespaceName n
         ns = Namespace nsn usrG adminG
+    chargeGasArgs info $ GRead $ sizeOf SizeOfV0 n
```
jmcardon, line 1700: Ditto here, why do we charge `GRead` on `n` when it's just a string?

On `existing` it's fine.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1601,9 +1676,11 @@ coreNamespace info b cont handler env = \case
     if T.null n then do
       (esLoaded . loNamespace) .== Nothing
       returnCEKValue cont handler (VString "Namespace reset to root")
-    else
+    else do
+      chargeGasArgs info $ GRead $ sizeOf SizeOfV0 n
```
jmcardon, line 1680: charging `GRead` on `n`. Needs to be on `ns` only

pact/Pact/Core/Gas/TableGasModel.hs
```diff
@@ -215,6 +248,43 @@ runTableModel = \case
...
      linearGasFactor = 38_000
   GModuleMemory bytes -> moduleMemoryCost bytes
+  GStrOp op -> case op of
+    StrOpLength len ->
+      let charsPerMg = 100
+      in MilliGas $ fromIntegral $ len `quot` charsPerMg + 1
+    StrOpConvToInt len ->
+      let mgPerChar = 20
+      in MilliGas $ fromIntegral $ len * mgPerChar + 1
+    StrOpParse len ->
```
jmcardon, line 258: This one is definitely interesting. 

Doesn't this highly depend on the parser used?

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -907,15 +936,18 @@ coreReadDecimal :: (CEKEval step b i m, MonadEval b i m) => NativeFunction step
 coreReadDecimal info b cont handler _env = \case
   [VString s] -> do
     viewEvalEnv eeMsgBody >>= \case
-      PObject envData ->
+      PObject envData -> do
+        chargeGasArgs info $ GObjOp $ ObjOpLookup s $ M.size envData
```
jmcardon, line 940: Hmm... 

This one I'm not sure whether we need to charge for this.

Maybe we do, but realistically, people pay for env data size in the payload, so I think we can forego charging simply for the object access here. Flat gas should be fine.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -907,15 +936,18 @@ coreReadDecimal :: (CEKEval step b i m, MonadEval b i m) => NativeFunction step
...
         case M.lookup (Field s) envData of
           Just (PDecimal p) -> returnCEKValue cont handler (VDecimal p)
           -- See [Note: Parsed Decimal]
           Just (PInteger i) -> returnCEKValue cont handler (VDecimal (Decimal 0 i))
-          Just (PString raw) -> case parseNumLiteral raw of
-            Just (LInteger i) -> returnCEKValue cont handler (VDecimal (Decimal 0 i))
-            Just (LDecimal l) -> returnCEKValue cont handler (VDecimal l)
-            _ -> returnCEK cont handler (VError "read-decimal failure" info)
+          Just (PString raw) -> do
+            chargeGasArgs info $ GStrOp $ StrOpLength $ T.length raw
```
jmcardon, line 946: Should this not charge on `StrOpParse`?

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -924,29 +956,38 @@ coreReadString :: (CEKEval step b i m, MonadEval b i m) => NativeFunction step b
 coreReadString info b cont handler _env = \case
   [VString s] -> do
     viewEvalEnv eeMsgBody >>= \case
-      PObject envData ->
+      PObject envData -> do
+        chargeGasArgs info $ GObjOp $ ObjOpLookup s $ M.size envData
```
jmcardon, line 960: Ditto, maybe we don't need to charge simply for reading a field from env data. Parsing is fine, env data can't ever grow to thousands and thousands of fields.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -924,29 +956,38 @@ coreReadString :: (CEKEval step b i m, MonadEval b i m) => NativeFunction step b
...
+readKeyset' info ksn = do
   viewEvalEnv eeMsgBody >>= \case
-    PObject envData ->
+    PObject envData -> do
+      chargeGasArgs info $ GObjOp $ ObjOpLookup ksn $ M.size envData
       case M.lookup (Field ksn) envData of
         Just (PGuard (GKeyset ks)) -> pure (Just ks)
-        Just (PObject dat) -> parseObj dat
+        Just (PObject dat) -> do
+          chargeGasArgs info $ GObjOp $ ObjOpLookup "keys" objSize
```
jmcardon, line 975: Ditto here: we're charging gas too often I think.

pact/Pact/Core/Gas/TableGasModel.hs
```diff
@@ -157,6 +156,35 @@ intShiftCost _ !rop
   | rop > 0 = MilliGas $ fromIntegral $ rop `quot` 1000
   | otherwise = MilliGas 0
 
+intPowCost :: Integer -> Integer -> MilliGas
+intPowCost !base !power
```
jmcardon, line 160: The gas for this seems incredibly complicated, and I've yet to test this locally.

What do you think about an approach such as [this one](https://github.com/kadena-io/pact/blob/master/src/Pact/Native/Ops.hs#L141)? It essentially simply charges on the actual operations that occur during the algorithm for integer exponentiation, which is simply a series of multiplications.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1521,13 +1587,18 @@ coreCompose info b cont handler env = \case
...
 
-createPrincipalForGuard :: (MonadEval b i m) => i -> Guard QualifiedName PactValue -> m (Pr.Principal)
+createPrincipalForGuard :: (MonadEval b i m) => i -> Guard QualifiedName PactValue -> m Pr.Principal
 createPrincipalForGuard info = \case
   GKeyset (KeySet ks pf) -> case (toList ks, pf) of
     ([k], KeysAll)
       | ed25519HexFormat k -> Pr.K k <$ chargeGas 1_000
     (l, _) -> do
       h <- mkHash $ map (T.encodeUtf8 . _pubKey) l
+      case pf of
```
jmcardon, line 1597: This seems excessive unless we can actually cause a gas discrep here.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
jmcardon: I agree with this change.

That said: you already gassed `Show`. I think we could reuse your `Show` gas for this. After all, `format` is essentially just `Show` with string splicing.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1229,17 +1279,25 @@ coreIntToStr info b cont handler _env = \case
 
 coreStrToInt :: (CEKEval step b i m, MonadEval b i m) => NativeFunction step b i m
 coreStrToInt info b cont handler _env = \case
-  [VString s] ->
-    checkLen info s *> doBase info cont handler 10 s
+  [VString s] -> do
+    chargeGasArgs info $ GStrOp $ StrOpLength $ T.length s
```
jmcardon, line 1283: I'm curious what we're charging for here? Shouldn't we be charging for parsing?

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1229,17 +1279,25 @@ coreIntToStr info b cont handler _env = \case
...
 
 coreStrToIntBase :: (CEKEval step b i m, MonadEval b i m) => NativeFunction step b i m
 coreStrToIntBase info b cont handler _env = \case
   [VInteger base, VString s]
-    | base == 64 -> checkLen info s *> case decodeBase64UrlUnpadded $ T.encodeUtf8 s of
-        Left{} -> throwExecutionError info (DecodeError "invalid b64 encoding")
-        Right bs -> returnCEKValue cont handler $ VInteger (bsToInteger bs)
-    | base >= 2 && base <= 16 -> checkLen info s *> doBase info cont handler base s
+    | base == 64 -> do
+        chargeGasArgs info $ GStrOp $ StrOpLength $ T.length s
```
jmcardon, line 1292: Ditto, I believe we should be charging for parsing.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -1218,9 +1263,14 @@ coreIntToStr info b cont handler _env = \case
     | v < 0 ->
       returnCEK cont handler (VError "int-to-str error: cannot show negative integer" info)
     | base >= 2 && base <= 16 -> do
+      let strLen = 1 + Exts.I# (IntLog.integerLogBase# base v)
+      chargeGasArgs info $ GConcat $ TextConcat $ GasTextLength $ fromIntegral strLen
```
rsoeldner, line 1267: Should this be `TextConcat`?

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
rsoeldner: Should charge for an object with the name + two guards (user and admin one)

gasmodel/Pact/Core/GasModel/Utils.hs
```diff
@@ -322,21 +324,83 @@ runNativeBenchmark
...
+
+stManaged :: [ManagedCap QualifiedName PactValue] -> StMod
+stManaged manageds = Endo $ esCaps.csManaged .~ S.fromList manageds
+
+stAddDef :: Text -> Def Name Type CoreBuiltin () -> StMod
+stAddDef name dfn = Endo $ (esLoaded.loToplevel %~ M.insert name (fqn, dkind))
+                         . (esLoaded.loAllLoaded %~ M.insert fqn dfn)
+  where
+  fqn = mkGasModelFqn name
+  dkind = case dfn of
```
jmcardon, line 375: call `defKind` from `Pact.Core.IR.Term`

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
jmcardon: Resolved. I think this is actually okay now after talking to Georg.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
jmcardon: Resolved. I think this is okay.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: The namespace itself is only charged if it's found by name. Hence to prevent undercharging in case the namespace isn't found, it's easier and less error-prone to just charge for the mere fact of looking up the name in the DB. _After_ it's found, it's also charged for the namespace object with its guards.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: In fact, it's even better than that, it should be `StrOpConvToInt` which has precisely the required factor! Thanks for spotting this, addressed.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: ditto, updated

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: Yep, it more or less corresponds to the semantics of `TextConcat` (and is reasonably close to the benchmarked results), which is
```
data ConcatType
  = TextConcat !GasTextLength
  -- ^ Total final string length
...
```

I agree the `GConcat`/`TextConcat` name is a bit overstretched (as I believe one of the comments somewhere in the code mentions), but we can rename that separately.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: ditto, updated

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: I'm inclined to leave this in for safety until we have these limited in length.

pact/Pact/Core/Gas/TableGasModel.hs
0xd34df00d: I think we've resolved this offline, marking as such.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: This is also resolved offline I guess, leaving as is and resolving.

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
0xd34df00d: I think we discovered today that `show` doesn't show all that needs to be shown, so we'll first show all that once lay shrouded in obscurity (like lists), and then we'll see how/if it can be unified charging-wise.

PR #125, opened 2024-05-08
Legacy Bindings v2

jmcardon:
PR checklist:

* [ ] Benchmark regressions

---
  - https://app.asana.com/0/0/1206312252755870
  - https://app.asana.com/0/0/1207288685407537


pact-tests/Pact/Core/Test/LegacySerialiseTests.hs
```diff
@@ -8,21 +8,77 @@
...
+import Pact.Core.IR.Term
+
+tests :: IO TestTree
+tests = testGroup "Legacy Repl Tests" <$> legacyTests
+
+legacyTestDir :: String
+legacyTestDir = "pact-tests" </> "legacy-serial-tests"
+
+
+replTestFiles :: IO [(String, [FilePath], [FilePath])]
```
imalsogreg, line 37:
```suggestion
-- Load a set of test cases from the file system.
-- Returns a list of tuples of:
--    - Test case name
--    - v5 modules with `expect` calls (*.repl)
--    - Legacy-encoded version of the same module (*.json)
replTestFiles :: IO [(String, [FilePath], [FilePath])]
```

pact-tests/Pact/Core/Test/LegacySerialiseTests.hs
```diff
@@ -8,21 +8,77 @@
...
+    files <- getDirectoryContents testPath
+    let replFiles = filter (isExtensionOf "repl") files
+        modFiles =  filter (isExtensionOf "json") files
+    pure (p, replFiles, modFiles)
+
+legacyTests :: IO [TestTree]
+legacyTests = do
+  t <- replTestFiles
+  forM t $ \(p, repl, lm) -> do
+    lm' <- sequence <$> traverse (toModuleData p) lm
```
imalsogreg, line 51:
```suggestion
    -- Perform legacy decoding on all .json files for the test.
    lm' <- sequence <$> traverse (toModuleData p) lm
```

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -1518,7 +1520,8 @@ applyContToValue (ObjC env info currfield fs vs cont) handler v = do
 
 applyContToValue (EnforceErrorC info _) handler v = case v of
   VString err -> returnCEK Mt handler (VError err info)
-  _ -> failInvariant info "enforce function did not return a string"
+  -- TODO: remove
```
imalsogreg, line 1523: Leftover TODO?

pact/Pact/Core/IR/Eval/Runtime/Utils.hs
```diff
@@ -74,6 +74,8 @@ import Pact.Core.Gas
 import Pact.Core.Guards
 import Pact.Core.Capabilities
 
+import Debug.Trace
```
imalsogreg, line 77: Leftover Trace

pact/Pact/Core/IR/ModuleHashing.hs
```diff
@@ -224,7 +224,8 @@ encodeApp operator operands =
 
 encodeSchema :: Schema -> Builder
 encodeSchema (Schema n _sc) =
-  encodeQualName n
+  -- TODO: This changes the hash, is this fine @Jose?
```
imalsogreg, line 227: This function only  differs only on schemas with no `QualifiedName`, IIUC? (I'm not sure how often we encounter schemas with no `QualifiedName`, so I'm not sure if this change to the hash is ok).

pact/Pact/Core/IR/Term.hs
```diff
@@ -348,6 +351,9 @@ instance (Pretty name, Pretty builtin, Pretty ty) => Pretty (Term name ty builti
       parens ("try" <+> pretty te <+> pretty te')
     ObjectLit n _ ->
       braces (hsep $ punctuate "," $ fmap (\(f, t) -> pretty f <> ":" <> pretty t) n)
+    InlineValue{} ->
```
imalsogreg, line 354: Even though this variant is only used in legacy terms, is there a chance we could encounter one why pretty-printing a legacy value? If so' should we give this variant a normal `pretty` representation?

pact/Pact/Core/IR/Term.hs
```diff
@@ -561,6 +577,18 @@ traverseModuleTerm f m =
   (mDefs . traversed) (traverseDefTerm f) m
 
 
+traverseIfDefTerm
```
imalsogreg, line 580: Are all the cases of this traversal correct? Or do `IfDfun` and `IfDPact` potentially hold builtins? If those variants can contain builtins, does the traversal need to traverse into them, in the same way it traverses into `IfDConst`?

I could be mixed up about interfaces (maybe they never contain code and therefore don't need to be traversed?)

pact/Pact/Core/IR/Eval/CEK.hs
jmcardon: Nope! Good catch. I do plan to cherry pick some changes onto master though

pact/Pact/Core/IR/ModuleHashing.hs
jmcardon: This entire function needs to go anyway. We need to restore source text hashing regardless.

pact/Pact/Core/IR/Term.hs
jmcardon: Yes, I agree I think we should

pact/Pact/Core/IR/Term.hs
jmcardon: Only `IfDConst` constains terms.

pact/Pact/Core/Type.hs
```diff
@@ -117,7 +117,7 @@ data Type
 instance NFData Type
 
 data Schema
-  = Schema QualifiedName (Map Field Type)
+  = Schema (Maybe QualifiedName) (Map Field Type)
```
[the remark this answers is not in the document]
rsoeldner: The reason is, that `TSchema` holds a `_tModule :: Maybe ModuleName`, which is required to construct a `QualifiedName`.

pact/Pact/Core/IR/ModuleHashing.hs
rsoeldner: @imalsogreg this is correct. I do think this is the behaviour we want in this particular case OR fail in case we hit a case with no `ModuleName`.

pact/Pact/Core/IR/Term.hs
rsoeldner: @imalsogreg Please feel free to re-ask :)

.github/workflows/nix.yml
```diff
@@ -40,7 +40,7 @@ jobs:
     - name: Build and cache artifacts
       run: |
         echo Building the project and its devShell
-        nix build .#check --log-lines 500 --show-trace
+        nix build .?submodules=1#check --log-lines 500 --show-trace
```
rsoeldner, line 43: Note, this is required for CI to fetch the new marmalade tests (which are supplied via git submodules).

pact-tests/pact-tests/coin-v5.repl
```diff
@@ -126,12 +126,12 @@
 ; w/o capability
 (expect-failure
   "direct call to credit fails"
-  "cap not in scope"
+  "require-capability: not granted: "
```
rsoeldner, line 129: Note, we had t correct our output to match legacy pact, as they are used inside the our test suite (marmalade v2) which we include via submodules.

pact/Pact/Core/Repl/Runtime/ReplBuiltin.hs
```diff
@@ -373,15 +375,17 @@ envExecConfig :: ReplCEKEval step => NativeFunction step ReplCoreBuiltin SpanInf
 envExecConfig info b cont handler _env = \case
   [VList s] -> do
     s' <- traverse go (V.toList s)
-    replEvalEnv . eeFlags .= S.fromList s'
-    let reps = PString . flagRep <$> s'
+    let (knownFlags, _unkownFlags) = partitionEithers s'
+    -- TODO: Emit warnings of unkown flags
```
rsoeldner, line 379: @jmcardon I think we should address this in a separate PR

pact-tests/Pact/Core/Test/LegacySerialiseTests.hs
```diff
@@ -8,21 +8,84 @@
...
+
+legacyTests :: IO [TestTree]
+legacyTests = do
+  t <- replTestFiles
+  forM t $ \(p, repl, lm) -> do
+    -- Perform legacy decoding on all `.json` files for that test
+    lm' <- sequence <$> traverse (toModuleData p) lm
+
+    case lm' of
+      Nothing -> error "Reading existing modules failed"
```
emilypi, line 60: Honestly this error is probably fine since we're failing in test prep. However, it does imply that there is a potential failure in the parsing and conversion to module data. If it's not too much  work, we should raise that exception specifically

pact-tests/Pact/Core/Test/LegacySerialiseTests.hs
```diff
@@ -8,21 +8,84 @@
...
+      Just ms -> do
+        pdb <- mockPactDb serialisePact_repl_spaninfo
+
+        -- add default spaninfo
+        let ms' = (fmap.fmap) (const def) ms
```
emilypi, line 65: should this be its own functionality? Seems really useful actually. Could be abstracted in terms of `SpanInfo`-classy optics

pact/Pact/Core/Hash.hs
```diff
@@ -29,6 +29,8 @@ module Pact.Core.Hash
 , placeholderHash
 , moduleHashToText
 , parseModuleHash
+  -- unsafe 
+, unsafeBsToModuleHash
```
emilypi, line 33: if it's unsafe, it should be documented why

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -2022,10 +2026,15 @@ isKeysetInSigs info cont handler env (KeySet kskeys ksPred) = do
   where
   matchKey k _ = k `elem` kskeys
   atLeast t m = m >= t
+  elide pk | T.length pk < 8 = pk
+           | otherwise = T.take 8 pk <> "..."
```
emilypi, line 2030: is the first case ever possible?

pact/Pact/Core/IR/Eval/Runtime/Utils.hs
```diff
@@ -96,7 +97,7 @@ lookupFqName fqn =
 getDefCap :: (MonadEval b i m) => i -> FullyQualifiedName -> m (EvalDefCap b i)
 getDefCap info fqn = lookupFqName fqn >>= \case
   Just (DCap d) -> pure d
-  _ -> failInvariant info "Expected DefCap"
+  d -> failInvariant info $ "Expected DefCap " <> renderFullyQualName fqn <> T.pack (show d)
```
emilypi, line 100: good old `tshow`. What would we be without it?

pact/Pact/Core/Repl/Runtime/ReplBuiltin.hs
emilypi: I agree here. This is actually kind of an odd piece of code. I _think_ we're just accumulating flags that make sense, but that begs the question of why this isn't just accumulating instead of lifting what we don't want into rights and what we want into lefts. I think we should error out if there's unknown flags tbqh

pact/Pact/Core/Serialise/CBOR_V1.hs
```diff
@@ -11,6 +11,9 @@ module Pact.Core.Serialise.CBOR_V1
   , encodeDefPactExec, decodeDefPactExec
   , encodeNamespace, decodeNamespace
   , encodeRowData, decodeRowData
+  -- used for
```
emilypi, line 14: used for?

pact/Pact/Core/Serialise/LegacyPact.hs
```diff
@@ -1,325 +1,1084 @@
 -- |
 {-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE PartialTypeSignatures  #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
-{-# LANGUAGE InstanceSigs #-}
+
```
emilypi, line 5: The whole thing needs docs

pact/Pact/Core/Serialise/LegacyPact.hs
```diff
@@ -1,325 +1,1084 @@
...
+runTranslateM :: TranslateM a -> Either String a
+runTranslateM a = runExcept (evalStateT (runReaderT a 0) [])
 
 decodeModuleData :: ByteString -> Maybe (ModuleData CoreBuiltin ())
-decodeModuleData = JD.decodeStrict'
+decodeModuleData bs = do
+  obj <- JD.decodeStrict' bs
+  either (const Nothing) Just (runTranslateM (fromLegacyModuleData obj))
+
+
```
emilypi, line 86:
```suggestion
```

tools/LegacyPactDbCheck.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE OverloadedStrings #-}
+
+-- |
+
```
emilypi, line 4: docs

emilypi (commented):
First Pass

pact/Pact/Core/IR/Eval/CEK.hs
rsoeldner: I was wondering about this too. Its a direct port of legacy pact.

pact/Pact/Core/Serialise/LegacyPact.hs
```diff
@@ -1,325 +1,1084 @@
...
-
-
-guardToPactValue :: Guard QualifiedName LegacyPactValue -> Guard QualifiedName PactValue
-guardToPactValue = \case
-  (GKeyset ks) -> GKeyset ks
-  (GKeySetRef kref) -> GKeySetRef kref
-  (GUserGuard (UserGuard n tm)) -> GUserGuard (UserGuard n (fromLegacyPactValue <$> tm))
-  (GCapabilityGuard (CapabilityGuard n args i)) -> GCapabilityGuard (CapabilityGuard n (fromLegacyPactValue <$> args) i)
-  (GModuleGuard mg) -> GModuleGuard mg
-  (GDefPactGuard dpg) -> GDefPactGuard dpg
```
emilypi, line 325: Do we have a canonical way that we're throwing invariants here, and where do they get picked up and communicated to the user by the interpreter? Some of them are inconsistent formats.

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+
+-- data Separator = Colon|ColonEquals|Comma deriving (Eq,Ord,Generic,Bounded,Enum,Show)
+
+
+-- instance JD.FromJSON Separator where
+--   parseJSON = withText "Separator" $ \t -> case t of
+--     ":" -> pure Colon
+--     ":=" -> pure ColonEquals
+--     "," -> pure Comma
+--     _ -> fail "Invalid separator"
```
emilypi, line 267: Lots of commented instances. Are these not needed?

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+import Bound.Scope
+import Bound.Var
+import Bound
+import Data.Maybe (fromMaybe)
+import qualified Data.Vector as V
+
+import qualified Data.Attoparsec.Text as AP
+import Data.Decimal (Decimal)
+import Pact.Time
+--import Control.Applicative ((<|>))
```
jmcardon, line 43: Remove

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+instance JD.FromJSON g => JD.FromJSON (ModuleDef g) where
+  -- parseJSON v = MDModule <$> parseJSON v <|> MDInterface <$> parseJSON v
+  parseJSON v
+    = MDModule <$> JD.parseJSON v
+    <|> MDInterface <$> JD.parseJSON v
+
+data Interface = Interface
+  { _interfaceName :: !ModuleName
+--  , _interfaceCode :: !Code
+--  , _interfaceMeta :: !Meta
```
jmcardon, line 98: if we don't need the field, can we just delete?

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+--  parseJSON = JD.lensyParseJSON 10
+  parseJSON = JD.withObject "Interface" $ \o ->
+    Interface
+      <$> o JD..: "name"
+      <*> o JD..: "imports"
+
+data Module g = Module
+  { _mName :: !ModuleName
+  , _mGovernance :: !(Governance g)
+--  , _mMeta :: !Meta
```
jmcardon, line 113: Ditto, delete unused fields?

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+  parseJSON v =
+    newKs v <|> oldKs v
+    where
+    oldKs = JD.withText "KeySetName" (pure . (`KeySetName` Nothing))
+    newKs =
+      JD.withObject "KeySetName" $ \o -> KeySetName
+        <$> o JD..: "ksn"
+        <*> (fromMaybe Nothing <$> o JD..:? "ns")
+
+-- data Meta = Meta
```
jmcardon, line 183: Delete commented code

pact/Pact/Core/Serialise/CBOR_V1.hs
jmcardon: Agree, either this needs to be deleted, or fixed.

pact/Pact/Core/Type.hs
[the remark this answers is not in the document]
jmcardon: Hmm, I think we can fail with an invariant failure if we ever see a `DefSchema` that doesn't have a `QualifiedName` along with it, and in the case that the defschema is inlined into the object type, we ignore this field. I think we should revert this.

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+      parseMR = JD.withObject "tagged ModRef" $ \o -> ModRef
+          <$> o JD..: "refName"
+          <*> o JD..: "refSpec"
+--          <*> pure def
+
+
+
+instance JD.FromJSON n => JD.FromJSON (Term n) where
+
+  -- "info" and "i" may be optional, so we don't consider those for matching
```
jmcardon, line 1675: we can remove these, unless it's necessary.

pact/Pact/Core/Serialise/LegacyPact/Types.hs
```diff
@@ -0,0 +1,1821 @@
...
+--    , _tInfo :: !Info
+    } |
+    TObject {
+      _tObject :: !(Object n)
+--    , _tInfo :: !Info
+    } |
+    TSchema {
+      _tSchemaName :: !TypeName
+    , _tModule :: !(Maybe ModuleName)
+--    , _tMeta :: !Meta
```
jmcardon, line 823: Clean up commented fields.

pact/Pact/Core/Repl/Runtime/ReplBuiltin.hs
rsoeldner: Im not sure about this one: silently ignoring only specific flags doesn't seem right to me. I'm in favour of ignoring them all for now and adding warnings to it. @emilypi wdyt?

pact/Pact/Core/Serialise/LegacyPact/Types.hs
rsoeldner: Cleaned them up

pact/Pact/Core/Serialise/LegacyPact.hs
```diff
@@ -1,325 +1,1084 @@
 -- |
 {-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE PartialTypeSignatures  #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
```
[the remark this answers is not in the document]
rsoeldner: remote it directly

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -2022,10 +2026,15 @@ isKeysetInSigs info cont handler env (KeySet kskeys ksPred) = do
...
   count = S.size kskeys
   run p matched =
     if p count matched then returnCEKValue cont handler (VBool True)
-    else returnCEK cont handler (VError "Keyset enforce failure" info)
+    else let
+      errMsg = "Keyset failure (" <> predicateToText ksPred <> "): "
```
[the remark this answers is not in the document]
rsoeldner: I like this but it would break our tests:
```
   marmalade-runner.repl:                                                   FAIL (0.04s)
        pact-tests/Pact/Core/Test/ReplTests.hs:111:
        marmalade-runner.repl:279:2: Program encountered an unhandled raised error: FAILURE: offer fails if OFFER capability is not signed by seller: expected error message 'Keyset failure (keys-all): [e4c6807d...]', got 'Keyset failure'        
```
as the marmalade expectations are very precise, see here: https://github.com/kadena-io/marmalade/blob/3eaaf08cd71304f43bc87ff97a0b29f776ef253b/pact/marmalade.repl#L280

My proposal would be we keep it as is for the time being and address this in a subsequent PR. 
What do you think?

0xd34df00d (approved):

PR #126, opened 2024-05-14
Improve callstack errors, check for recursion.

jmcardon:
This PR improves the state of error messages in core, as well as adds a runtime recursion check for module references.

Given this repl file:
```
(module caller g

  (defcap g () true)

  (defun caller ()
    (caller1)
  )

  (defun caller1 ()
    (caller2 31)
  )

  (defun caller2 (a:integer)
    (caller3 31 "bob`")
  )

  (defun caller3 (a:integer arg:string)
    (enforce false "whoopsie!")
  )
  )

(caller)
```

Pre-this PR, the following is a repl snippet of the error emitted by loading the file:
```
pact>(load "example.repl" true)
"Loading example.repl..."
Loaded module caller, hash Z9GxDb9xe8joF3x1SfeqbMuN9RrSSekRm5N2TBTJ5zI
example.repl:18:4: Program encountered an unhandled raised error: whoopsie!
 18 |     (enforce false "whoopsie!")
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

Post-this, the error improves to

```
pact>(load "example.repl" true)
"Loading example.repl..."
Loaded module caller, hash Z9GxDb9xe8joF3x1SfeqbMuN9RrSSekRm5N2TBTJ5zI
example.repl:18:4: Program encountered an unhandled raised error: whoopsie!
 18 |     (enforce false "whoopsie!")
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  at (caller.caller3 31 "bob`")
  at (caller.caller2 31)
  at (caller.caller1)
  at (caller.caller)
  ```

PR checklist:

### DIFF
--- a/pact-tests/pact-tests/modref-recursion.repl
+++ b/pact-tests/pact-tests/modref-recursion.repl
@@ -1,7 +1,7 @@
 
 (interface call
 
-  (defun callF (m:module{call}))
+  (defun callF:integer (m:module{call}))
   )
 
 (module knot1 g
@@ -9,7 +9,7 @@
 
   (implements call)
 
-  (defun callF (m:module{call})
+  (defun callF:integer (m:module{call})
     (+ 1 2) ; call something so it costs gas
     (m::callF knot1)
   )
@@ -20,7 +20,7 @@
 
   (implements call)
 
-  (defun callF (m:module{call})
+  (defun callF:integer (m:module{call})
     (+ 1 2) ; call something so it costs gas
     (m::callF knot2)
   )
@@ -29,4 +29,4 @@
 (env-gasmodel "table")
 (env-gaslimit 10) ; ensures test does not run forever in case recursion breaks
 
-(expect-failure "Recursion should fail @ runtime" "asdf" (knot2.callF knot1))
+(expect-failure "Recursion should fail @ runtime" (knot2.callF knot1))
--- a/pact-tests/pact-tests/modref-recursion.repl
+++ b/pact-tests/pact-tests/modref-recursion.repl
@@ -1,0 +1,32 @@
+
+(interface call
+
+  (defun callF (m:module{call}))
+  )
+
+(module knot1 g
+  (defcap g () true)
+
+  (implements call)
+
+  (defun callF (m:module{call})
+    (+ 1 2) ; call something so it costs gas
+    (m::callF knot1)
+  )
+  )
+
+(module knot2 g
+  (defcap g () true)
+
+  (implements call)
+
+  (defun callF (m:module{call})
+    (+ 1 2) ; call something so it costs gas
+    (m::callF knot2)
+  )
+  )
+
+(env-gasmodel "table")
+(env-gaslimit 10) ; ensures test does not run forever in case recursion breaks
+
+(expect-failure "Recursion should fail @ runtime" "asdf" (knot2.callF knot1))
--- a/pact/Pact/Core/IR/Eval/CEK.hs
+++ b/pact/Pact/Core/IR/Eval/CEK.hs
@@ -326,7 +326,7 @@ mkDefunClosure d fqn e = case _dfunTerm d of
   Nullary body i ->
     pure (Closure fqn NullaryClosure 0 body (_dfunRType d) e i)
   _ ->
-    failInvariant di ("definition is not a closure: " <> T.pack (show d))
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

File: pact/Pact/Core/IR/Eval/CEK.hs [hunk 1]
```diff
@@ -731,30 +730,41 @@ acquireModuleAdmin i cont handler env mdl = do
...
   -> m (Cont step b i m)
 pushStackFrame info cont mty sf = do
+  checkRecursion
   esStack %== (sf :)
   pure (StackPopC info mty cont)
+  where
+  checkRecursion = do
+    RecursionCheck currentCalled <- usesEvalState esCheckRecursion NE.head
+    let qn = fqnToQualName (_sfName sf)
+    when (S.member qn currentCalled) $ throwExecutionError info (RuntimeRecursionDetected qn)
```
0xd34df00d, line 764: At first I was a little wary that we forbid _any_ kind of mutual recursion, but then I figured that we're already going to great lengths to forbid all kind of possibly well-founded (but possibly not!) recursion cases, so that's OK.

0xd34df00d (approved):

pact/Pact/Core/IR/Eval/CEK.hs
jmcardon: 👍 

This is just reinstating prod behavior, just slightly more efficiently using sets instead of re-traversing the callstack.

pact/Pact/Core/IR/Eval/Runtime/Types.hs
```diff
@@ -594,16 +593,16 @@ data EvalCapType
...
 
 -- | State to preserve in the error handler
-data ErrorState
-  = ErrorState (CapState QualifiedName PactValue) [StackFrame]
+data ErrorState i
+  = ErrorState (CapState QualifiedName PactValue) [StackFrame i]
   deriving (Show, Generic)
 
-instance NFData ErrorState
+instance NFData i => NFData (ErrorState i)
```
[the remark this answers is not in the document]
jmcardon: I actually kind of like standalone deriving instances for some reason :D

pact-tests/pact-tests/modref-recursion.repl
```diff
@@ -0,0 +1,32 @@
...
+  (defun callF:integer (m:module{call})
+    (+ 1 2) ; call something so it costs gas
+    (m::callF knot2)
+  )
+  )
+
+(env-gasmodel "table")
+(env-gaslimit 10) ; ensures test does not run forever in case recursion breaks
+
+(expect-failure "Recursion should fail @ runtime" (knot2.callF knot1))
```
emilypi, line 32: let's add a case for direct self-recursion as well. Also, is mutual recursion covered for both modrefs and two concrete mutually recursive calls?

pact/Pact/Core/IR/Eval/CEK.hs
emilypi, line 764: So correct me if i'm wrong, but the process is as follows: 

When we see a fqn that's new, we check the current frame calling context to see if there's already a reference to it anywhere, treating the `NonEmpty RecursionCheck` state as a second stack that tracks names that it's already seen, and if it has been seen already, abort, but if not, load it into the associated recursion tracking context if it's not. 

This could get pretty costly. We'll need to find a way to gas this because any significantly complicated `defun` is going to have a lot of recursion checks no?

0xd34df00d: My gut feel is this shouldn't be too costly relative to the other stuff CEK machine is doing, but it'd make a great benchmark case! I can take a look at that separately soon-ish.

jmcardon: > This could get pretty costly.

Why would this be costly? The set will not grow very large in practice, and it just checks a qualified name. 

> We'll need to find a way to gas this because any significantly complicated defun is going to have a lot of recursion checks no?

We already charge an overhead for function application. The overhead covers the cost of a few things anyway:
- Pushing the arguments onto the environment
- Pushing a stackframe into the context stack
- Recursion checking.

In practice, that set will be small, ankd even if it were 10000 functions deep, it would not be slow enough for it to matter in the grand scheme of things, I believe, but we will simply use the benchmarks for function app to verify this 👍

PR #128, opened 2024-05-15
incremental gassed serialization

imalsogreg:
Add gas costing to serialization of PactValues.

This is primarily a change to the (1) `PactSerialise` struct, which specifies how to serialize and deserialize things like `RowData` for user table reads/writes, `KeySet`s and `Module`s, for persisting guards, defpact bodies, etc., and (2) `Serialize/CBOR_V1.hs`, which now tracks gas usage while serializing `RowData` and `PactValue`.

Specifically the field `PactSerialise::_encodeRowData` changed its type from `RowData -> ByteString` to `RowData -> GasM (PactError i) ByteString`, to indicate that serialization of `RowData` is monadic, it runs in a new monad called `GasM` that tracks gas usage and will fail if too much gas is used.

This change has several knock-on effects:
 - The `Persistence` struct, which determines how values are persisted, and which relies on `PactSerialize` got updates for all fields that do encoding (adding the `GasM` monad).
 - The `Persistence` module and the `Gas` module each had to be broken out into `Types.hs` and `Utils.hs`, to break cyclic dependencies that arose when mentioning `Gas` in the context of `Persistence`.
 - Operations that used to be free now cost some gas, so tests of gas costs got their costs tweaked as needed.

Determining the gas constants to use in `CBOR_V1.hs` required a lot of benchmarking. The benchmarks are collected in a standalone executable `Bench.hs`, and the comments in that file explain how we converted the results into gas coefficients used during serialization.

PR checklist:

* [ ] Benchmark regressions


pact/Pact/Core/Gas.hs
```diff
@@ -246,15 +231,44 @@ constantGasModel unitPrice gl
...
-gasToMilliGas (Gas n) = MilliGas (n * millisPerGas)
-{-# INLINE gasToMilliGas #-}
+runGasM
+  :: i
+  -> GasMEnv
+  -> GasM (PactError i) a
+  -> IO (Either (PactError i) a)
+runGasM info env (GasM m) = do
+  res <- try $ runExceptT $ runReaderT m env
+  let fakeStackFrame = []
```
jmcardon, line 259: if we make a version of `runGasM` that works inside of `MonadEval`, we should be able to have a version of this combinator that throws with a stack frame :)

`runGasMEval` maybe?

pact/Pact/Core/Gas.hs
```diff
@@ -246,15 +231,44 @@ constantGasModel unitPrice gl
   }
 
 freeGasModel :: GasModel b
-freeGasModel = constantGasModel mempty (MilliGasLimit mempty)
+freeGasModel = constantGasModel mempty (MilliGasLimit (MilliGas maxBound)) -- TODO: some tests seem to charge gas even with freeGasModel.
```
jmcardon, line 234: Is this because we're charging gas directly?

Maybe GasM needs to account for the "free gas model"...

Maybe the free gas model should default to a massive gas limit instead?

pact/Pact/Core/SizeOf.hs
```diff
@@ -79,21 +88,47 @@ import Pact.Core.Namespace (Namespace)
...
+--   For example, we have found the SizeOfV0 measurements to be sufficient
+--   when computing the size of a module during module-loading, so the
+--   code that charges gas for module load hardcodes SizeOfVersion to
+--   SizeOfV0. Size estimates for writing to user tables change more
+--   frequenly, so the gas computation for that refers to pact version
+--   flags to get the appropriate SizeOfVersion.
 data SizeOfVersion
-  = SizeOfV0
-  deriving Show
+  = SizeOfV2
```
jmcardon, line 114: We should only have `SizeOfV0` I think. For now. The prior versions are purely historical to prod.

pact/Pact/Core/SizeOf.hs
```diff
@@ -164,82 +212,97 @@ instance SizeOf SBS.ShortByteString where
...
-    GHC.IP bn -> fromIntegral $ I# (GHC.wordArraySize# bn)
-    GHC.IN bn -> fromIntegral $ I# (GHC.wordArraySize# bn)
-
-
-instance SizeOf Word64 where
-  sizeOf _ _ = 2 * wordSize64
+  sizeOf ver i = countBytes ver $ case ver of
+    SizeOfV0 ->
+      if i < 0 then 0 else ceiling ((logBase 100000 (realToFrac i)) :: Double)
+    _ ->
```
jmcardon, line 222: Let's only use the bottom case.

pact/Pact/Core/SizeOf.hs
```diff
@@ -367,11 +452,23 @@ instance (SizeOf n, SizeOf t, SizeOf b, SizeOf i) => SizeOf (DefPact n t b i)
...
 instance (SizeOf n, SizeOf t, SizeOf i) => SizeOf (IfDefCap n t i)
 instance (SizeOf n, SizeOf t, SizeOf b, SizeOf i) => SizeOf (IfDef n t b i)
 instance (SizeOf n, SizeOf t, SizeOf b, SizeOf i) => SizeOf (Interface n t b i)
 
 instance SizeOf Namespace
+
+
+type Bytes = Word64
+
+newtype SizeOfByteLimit
```
jmcardon, line 469: Can we move these closer to the top of the file? These definitions are pretty important.

pact/Pact/Core/Serialise/CBOR_V1.hs
```diff
@@ -76,8 +90,67 @@ encodeNamespace = toStrictByteString . encode
 decodeNamespace :: ByteString -> Maybe Namespace
 decodeNamespace bs =either (const Nothing) (Just . snd) (deserialiseFromBytes decode (fromStrict bs))
 
-encodeRowData :: RowData -> ByteString
-encodeRowData = toStrictByteString . encode
+-- TODO: Do we want a gas log entry here???
+-- most likely not, maybe just pre/post?
```
jmcardon, line 94: Great question! I doubt it. I think just for the post-quantity.

pact/Pact/Core/Serialise/CBOR_V1.hs
```diff
@@ -76,8 +90,67 @@ encodeNamespace = toStrictByteString . encode
...
+      LString s ->
+        -- See the analysis in `Bench.hs` - `pact-string-2` for details.
+        chargeGasMSerialize info $ MilliGas $ 1 + 69 * fromIntegral (Text.length s) `div` 1000
+      LInteger i ->
+        -- See the analysis in `Bench.hs` - `pact-ineger-2` for details.
+        chargeGasMSerialize info $ MilliGas $ 2 * fromIntegral (I# (IntLog.integerLogBase# 10 (abs i)))
+      LDecimal d ->
+        chargeGasMSerialize info $ MilliGas $ 59 + 2 * fromIntegral (I# (IntLog.integerLogBase# 10 (decimalMantissa d)))
+      LBool _ -> chargeGasMSerialize info $ MilliGas 52
+      LUnit -> chargeGasMSerialize info $ MilliGas 51
```
jmcardon, line 152: I see what you mean about the constants. 

Maybe a `Pact.Gas.Constants` is necessary. There's definitely quite a lot here in this file.

pact/Pact/Core/Persistence/Utils.hs
```diff
@@ -0,0 +1,69 @@
...
+-- failIfUsesGas :: MilliGas -> m ()
+-- failIfUsesGas _ = error "Expected no gas use (even charges of 0 gas)"
+
+
+dbOpDisallowed :: MonadIO m => m a
+dbOpDisallowed = liftIO $ putStrLn "OpDisallowed" >> throwIO OpDisallowed
+
+dbOpDisallowed2 :: forall i m a. (MonadError (PactError i) m, MonadIO m, Default i) => m a
+dbOpDisallowed2 = do
+  liftIO (putStrLn "OpDisallowed")
```
jmcardon, line 60: Remove 👀

pact/Pact/Core/Persistence/Types.hs
```diff
@@ -0,0 +1,250 @@
...
+ , HasLoaded(..)
+ , HasPactDb(..)
+ , Domain(..)
+ , WriteType(..)
+ , Purity(..)
+ , RowData(..)
+ , ExecutionMode(..)
+ , mdModuleName
+ , mdModuleHash
+--  , readModule, writeModule
```
jmcardon, line 29: Need to clean up commented exports/imports.

pact/Pact/Core/Gas.hs
```diff
@@ -166,8 +149,10 @@ data GasArgs
   | GSearch !SearchType
   -- ^ Gas costs for searches
   | GPoseidonHashHackAChain !Int
-  -- ^ poseidon-hash-hack-a-chain costs
+  -- ^ poseidon-hash-hack-a-chain costs.
   | GModuleMemory !Word64
+  | GCountBytes
```
jmcardon, line 154: Missing the `NBytes` argument?

pact/Pact/Core/Gas.hs
```diff
@@ -121,7 +102,9 @@ data ZKGroup
   -- ^ Group one, that is Fq in Pairing
   | ZKG2
   -- ^ Group two, that is, Fq2 Pairing
-  deriving (Show, Generic, NFData)
+  deriving (Show, Generic)
+
+instance NFData ZKGroup
```
jmcardon, line 107: Is there a reason that all of the other types were not moved to `Pact.Gas.Types`?

Just a nitpick on our current core convention, but I think that our gas types should most likely be in one place, this includes gas args.

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -1414,7 +1415,7 @@ applyContToValue (BuiltinC env info frame cont) handler cv = do
             , (Field "key", PString key)
             , (Field "value", PObject rdata)]
       CreateTableC (TableValue tn _ _) -> do
-        liftDbFunction info (_pdbCreateUserTable pdb tn)
+        liftGasM info (_pdbCreateUserTable pdb info tn)
```
jmcardon, line 1418: 👍

pact/Pact/Core/SizeOf.hs
imalsogreg: Oh, this is resolved by your later comment.

pact/Pact/Core/Gas.hs
imalsogreg: This is just the cost of calling `countBytes`, which happens during (incremental) `SizeOf` computations. That cost doesn't depend on how many bytes get counted.

pact/Pact/Core/Gas.hs
imalsogreg: That's it, yes - `GasM` charges gas directly, rather than looking up the values from any `GasModel`.

A quick and dirty solution would be keeping the change currently in the PR - `freeGasModel` just has an infinite limit. But that's a hack.

Probably a better solution is for `GasM` to carry a `GasModel` in its `GasMEvn`, and for the serialization factors to be gas args, like all other gassed operations. WDYT @jmcardon ?

pact/Pact/Core/SizeOf.hs
imalsogreg: :+1:  Done

pact/Pact/Core/Gas.hs
imalsogreg: Thx, done

pact/Pact/Core/SizeOf.hs
imalsogreg: Done

pact/Pact/Core/Gas.hs
rsoeldner: I've derived `Bounded` for `MilliGasLimit` in  https://github.com/kadena-io/pact-5/blob/37633735c86f7eedaf55800f4670570eb2c88603/pact/Pact/Core/Gas.hs#L68 

In my opinion, we should make `freeGasModel :: GasModel b` consistent with `tableGasModel :: MilliGasLimit -> GasModel CoreBuiltin`, i.e. taking `MilliGasLimit`.

bench/Bench.hs
```diff
@@ -0,0 +1,1022 @@
...
+ \(module m G \
+ \  (defcap G () true) \
+ \  (defun foo()       \
+ \   \"return 10\"     \
+ \   10)               \
+ \)"
+
+
+
+coinModuleCode :: String
```
rsoeldner, line 366: I do prefer a dedicated file, why not `coin-v5.pact`?

pact-tng.cabal
```diff
@@ -97,6 +97,7 @@ common pact-common
     FlexibleInstances
     NumericUnderscores
     TypeOperators
+    PartialTypeSignatures
```
rsoeldner, line 100: I think we decided against it?

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -1556,9 +1561,8 @@ applyContToValue (DefPactStepC env cont) handler v =
           done = (not (_psRollback ps) && isLastStep) || _psRollback ps
         when (nestedPactsNotAdvanced pe ps) $
           throwExecutionError def (NestedDefpactsNotAdvanced (_peDefPactId pe))
-        liftDbFunction def
-          (writeDefPacts pdb Write (_psDefPactId ps)
-            (if done then Nothing else Just pe))
+        writeDefPacts stack def pdb Write (_psDefPactId ps) -- TODO: def used because we have no `info`.
```
rsoeldner, line 1564: How about adding info to the frame?

pact/Pact/Core/Persistence/Types.hs
```diff
@@ -0,0 +1,245 @@
...
+ , TxLog(..)
+ , toUserTable
+ , objectDataToRowData
+ , rowDataToObjectData
+ ) where
+
+import Control.Applicative((<|>))
+import Control.Lens
+-- import Control.Exception(throwIO)
+-- import qualified Control.Monad.Catch as Exceptions
```
rsoeldner, line 40: cleanup

pact/Pact/Core/IR/Eval/CEK.hs
imalsogreg: Beautiful idea! Done

bench/Bench.hs
imalsogreg: I've simply removed that benchmark. We don't really need to track it, and we can bring it back if needed. Removing it allowed me to delete this awkward copy-paste of coin-v5.repl

pact/Pact/Core/Gas.hs
imalsogreg: I added a table of serialization gas costs to `GasModel`, and added `GasModel` to `GasMEnv`. Now we can specify that serialization is free within the `freeGasModel`, and those costs get propagated through to the code that runs serialization.

jmcardon (approved):

0xd34df00d (approved):

PR #134, opened 2024-05-28
add initial gas golden

rsoeldner:
PR checklist:

* [ ] Benchmark regressions


emilypi (approved):

pact-tests/gas-goldens/resume.repl
```diff
@@ -0,0 +1,15 @@
+(module testmodule g
+  (defcap g () true)
+
+  (defschema schema result:string)
+
+  (defpact test ()
+    (step (let ((x:object{schema} {"result": "hi"})) (yield x)))
+    (step (resume {"result" := r} (+ r " Kadena")))))
+    
+
```
jmcardon, line 10: If this is a repl file could, could you `(env-gas 0)` before running the `(continue-pact 1)`? It will get rid of the noise from the other things being executed.

pact-tests/Pact/Core/Test/GasGolden.hs
```diff
@@ -0,0 +1,130 @@
...
+  , ("~", "binary-inverse")
+  ]
+
+fileNameToOp :: M.Map Text Text
+fileNameToOp = M.fromList [(v,k) | (k, v) <- M.toList opToFileName]
+
+runGasTest :: FilePath -> IO (Maybe MilliGas)
+runGasTest file = do
+  src <- T.readFile file
+  pdb <-  mockPactDb serialisePact_repl_spaninfo
```
0xd34df00d, line 109:
```suggestion
  pdb <- mockPactDb serialisePact_repl_spaninfo
```

pact-tests/Pact/Core/Test/GasGolden.hs
```diff
@@ -0,0 +1,130 @@
...
+            , _replCurrSource = source
+            , _replEvalEnv = ee'
+            , _replUserDocs = mempty
+            , _replTLDefPos = mempty
+            , _replTx = Nothing
+            , _replNativesEnabled = False
+            }
+  stateRef <- newIORef rstate
+  runReplT stateRef (interpretReplProgram source (const (pure ()))) >>= \case
+    Left _ -> pure Nothing
```
0xd34df00d, line 129: I found it helpful in my tests to output the error as well, so it might be worth it to return the error here and print it in `gasGoldenTests`.

pact-lsp/Pact/Core/LanguageServer.hs
```diff
@@ -205,7 +204,6 @@ sendDiagnostics nuri mv content = liftIO runPact >>= \case
     runPact = do
       let file = fromMaybe "<local>" $ uriToFilePath (fromNormalizedUri nuri)
       pdb <- mockPactDb serialisePact_repl_spaninfo
-      gasRef <- newIORef (Gas 0)
```
0xd34df00d, line 208: Just to confirm — we're using `eeGasRef` for accounting now, right?

0xd34df00d (approved):

pact-lsp/Pact/Core/LanguageServer.hs
rsoeldner: Yes

PR #135, opened 2024-05-29
tests: remove marmalade submodule

thoughtpolice:
This replaces the marmalade submodule with a simple script that can be run and will automatically commit an update for you.

We've had numerous discussions about submodules recently and to keep it short: this one isn't needed and in general just causes pains. Replacing it with an occasionally-updated snapshot that exists within the repo is much better and doesn't cause so many problems (e.g. needing to constantly `git submodule update`.)

This PR is large, but only because it commits a copy of marmalade in the 3rd commit. Otherwise, it's very small, and the most important thing is the script itself, which works on my amd64/Linux machine. It should also work fine on macOS, I think.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -40,7 +40,7 @@ jobs:
     - name: Build and cache artifacts
       run: |
         echo Building the project and its devShell
-        nix build .?submodules=1#check --log-lines 500 --show-trace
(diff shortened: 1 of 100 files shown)


jmcardon (approved):

rsoeldner (approved):
Thank you Austin 👍

PR #136, opened 2024-05-30
Equality tweaks

0xd34df00d:
PR checklist:

* [ ] Benchmark regressions


jmcardon (approved):

0xd34df00d:
Something funny's with macos-m1 runners, they don't get picked up and the job is stuck in queue. Merging as is since the other two runners pass.

PR #142, opened 2024-06-06
Show via `pretty` + tweaks

0xd34df00d:
PR checklist:

* [ ] Benchmark regressions


rsoeldner (approved):
nice work 👍

pact/Pact/Core/IR/Eval/CoreBuiltin.hs
```diff
@@ -385,24 +385,17 @@ rawSqrt info b cont handler _env = \case
     returnCEKValue cont handler (VLiteral (LDecimal (f2Dec result)))
   args -> argsError info b args
 
--- Todo: fix all show instances
+showPactValue :: MonadEval b i m => i -> PactValue -> m T.Text
+showPactValue info pv = do
+  sz <- sizeOf SizeOfV0 pv
+  chargeGasArgs info $ GConcat $ TextConcat $ GasTextLength $ fromIntegral sz
+  pure $ T.pack $ show $ Pretty.pretty pv
```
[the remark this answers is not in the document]
0xd34df00d: Cool! I've reused our `renderCompactText`.

pact-tests/gas-goldens/builtinGas.golden
```diff
@@ -96,7 +96,7 @@ round: 50
 scalar-mult: 360225
 select: 100383581
 shift: 4
-show: 101
+show: 4002
```
emilypi, line 99: wow that's a big jump

pact/Pact/Core/IR/Eval/Direct/Evaluator.hs
```diff
@@ -1763,24 +1763,15 @@ rawSqrt info b _env = \case
     return (VLiteral (LDecimal (f2Dec result)))
   args -> argsError info b args
 
--- Todo: fix all show instances
+showPactValue :: MonadEval b i m => i -> PactValue -> m T.Text
+showPactValue info pv = do
+  sz <- sizeOf SizeOfV0 pv
+  chargeGasArgs info $ GConcat $ TextConcat $ GasTextLength $ fromIntegral sz
+  pure $ Pretty.renderCompactText pv
```
emilypi, line 1770: Small bikesheddey nitpick: if i'm calling `showPactValue`, i don't assume any rendering or compaction: I assume it's using the show instance on pact values. Maybe we keep `renderPactValue` as a name, because it implies there's work being done on the value to produce text.

0xd34df00d: Good point, I agree the current name is slightly misleading.

PR #158, opened 2024-07-03
Review pact functions

lsgunnlsgunn:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/docs/builtins/Operators/and-q.md
+++ b/docs/builtins/Operators/and-q.md
@@ -1,10 +1,10 @@
 ## and?
 
-Use `and?` to apply a logical AND operation to the results of applying a `value` to functions `oper1` and `oper2`, with short-circuiting.
+Use `and?` to apply a logical AND operation to the results of applying a specified `value` to application functions `oper1` and `oper2`, with short-circuit evaluation.
 
 ### Basic syntax
 
-To apply a logical 'and' to the results of applying `value` to functions `oper1` and `oper2`, use the following syntax:
+To apply a logical AND operation to the results of applying a specified `value` to the functions `oper1` and `oper2`, use the following syntax:
 
 ```pact
 (and? oper1 oper2 value)
@@ -26,7 +26,7 @@ The `and?` function returns a boolean value based on the result of applying `val
 
 ### Examples
 
-The following example demonstrates the `and?` operation in the Pact REPL:
+The following example demonstrates how to use the `and?` function in the Pact REPL:
 
 ```pact
 pact> (and? (> 20) (> 10) 15)
--- a/docs/builtins/Operators/and-q.md
+++ b/docs/builtins/Operators/and-q.md
@@ -2,7 +2,7 @@
 
 Use `and?` to apply a logical AND operation to the results of applying a specified `value` to application functions `oper1` and `oper2`, with short-circuit evaluation.
 
-In practice, the data type of the `value` you specify is used to evaluate the expression in each function and to return a boolean value for the logical AND operation performed by the `and?` function:
+In practice, the data type of the `value` you specify is used to evaluate the expression in each function and to return the resulting boolean value for the logical AND operation performed by the `and?` function:
 
 (defun <a> and?:bool (oper1:(<a> -> bool) oper2:(<a> -> bool) value:<a>))
 
--- a/docs/builtins/Operators/and-q.md
+++ b/docs/builtins/Operators/and-q.md
@@ -6,7 +6,7 @@ You can use any data type for the `value` argument as long as the two functions 
 
 By convention, the data type <a> is used if an argument represents a type-bound parameter like the `value` argument in this function: 
 
-(defun <a> and?:bool (func1:(<a> -> bool) func2:(<a> -> bool) value:<a>))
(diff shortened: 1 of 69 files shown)


docs/builtins/Operators/div.md
```diff
@@ -1,37 +1,38 @@
-## /
-The `/` function divides the first argument `x` by the second argument `y`.
+## divide (/)
```
rsoeldner, line 1: hmm, not sure about this change but probably makes sense 👍

rsoeldner (approved):

docs/builtins/Operators/and-q.md
```diff
@@ -1,33 +1,38 @@
...
 ### Arguments
 
 Use the following arguments to specify the functions and `value` for the `and?` operation.
 
 | Argument | Type | Description |
 | --- | --- | --- |
-| `a` | `x:<r> -> bool` | Specifies the first function to apply `value` to. |
-| `b` | `x:<r> -> bool` | Specifies the second function to apply `value` to. |
-| `value` | `<r>` | Specifies the value to apply to both `a` and `b` functions. |
+| `oper1` | function | Specifies the first function to apply the specified `value` to. The result of applying the specified value returns a boolean value. |
```
jmcardon, line 19: I am unsure whether i like this. Maybe `function x:<r> -> bool`. 

The types simply aren't clear here.

docs/builtins/Operators/and-q.md
```diff
@@ -1,33 +1,38 @@
...
+| `oper2` | function | Specifies the second function to apply the specified `value` to. The result of applying the specified value returns a boolean value.|
+| `value` | any | Specifies the value to apply to both `oper1` and `oper2` functions. |
```
jmcardon, line 21: Ditto here, the value isn't `any`, although it will technically execute and fail at runtime, the type of the function and value are intertwined. That is to say: `oper1` and `oper2` will not accept any value `any`, the type of this function is `forall a. (a -> bool) -> (a -> bool) -> a -> bool`. So for any `a`, it applies `oper1` and `oper2` to the argument and treat it like `and (oper1 v) (oper2 v)`.

That's why the types used to be `<r>`, indicating some bound type variable, but our table format does not exactly make this easy to display.

jmcardon (changes requested):
I am not sure I like the overall changes from `<a>` to describe bound type variables to `any`, as well as the function type removals.

It unfortunately makes it very unclear what types are accepted by a particular function, or whether it expects a value of the same type.

If this is a bit hard to display in a table format, then likely we need to maybe rethink the entire table formatting structure here.

Take for example rust documentation (taken from [tokio](https://docs.rs/tokio/latest/tokio/io/struct.BufStream.html))
```
pub fn with_capacity(
        reader_capacity: usize,
        writer_capacity: usize,
        stream: RW,
    ) -> BufStream<RW> 
    
Creates a BufStream with the specified [BufReader](https://docs.rs/tokio/latest/tokio/io/struct.BufReader.html) capacity and [BufWriter](https://docs.rs/tokio/latest/tokio/io/struct.BufWriter.html) capacity.

See the documentation for those types and [BufStream](https://docs.rs/tokio/latest/tokio/io/struct.BufStream.html) for details.
```

This allows us to talk about both type arguments and type variables. So what if we take for example a function such as `and?` and change it to something like (and this is mostly just brainstorming):

```lisp
(defun <a> and?:bool (oper1:(<a> -> bool) oper2:(<a> -> bool) value:<a>)
```

where `(defun <a>` introduces the type variable binding `a` that then is used in the function signature arguments?

That way, we can display the signature clearly, and if we want better descriptions of the arguments, we can simply keep the table, but remove the `type` column.

docs/builtins/Operators/and-q.md
lsgunnlsgunn: There are a few options we can pursue.
If you want to use something like `x:<r> -> bool` as a type, we need to create a documentation conventions section that clearly explains what this means to an uninitiated person. Part of the problem in using stuff like x, <x>, <r>, and <a>  is that I often have no idea what it is trying to convey. 
I would prefer that we add more descriptive information about the functions and types expected (somewhat like the Rust example, I think).  I realize using `function` here is probably too much of a punt, but in trying to test the examples of the functions, I found some of the type description just too cryptic to be useful.

The two most likely fixes in my mind would be:
- Doc conventions with a few paragraphs explaining what the hieroglyphic notations are meant to convey that can be applied to all cases where they are used. 
- A Functions (or something like that) subsection where it's applicable that explains what's expects in the functions that are arguments for other functions.

I don't think there are a ton of these. But I did stumble over several like the and? that were hard for me to figure out.

docs/builtins/Operators/and-q.md
lsgunnlsgunn: Maybe instead of `any` we could use a wildcard like `*` and have a convention that where you see `*` (or `any` for that matter) it means that the type depends on the function or expression being evaluated. We could even use `<r>` if it consistently means a **bound type variable** and we document that it means that. I didn't find any documentation conventions that explained this, so I wasn't convinced that we had actual conventions consistently applied.

PR #173, opened 2024-07-11
fix pact version for CI

jmcardon:
This PR simply creates a static version of the autogenerated `PackageInfo_pact_tng`, while we fix our CI.


pact/Pact/Core/Version.hs
```diff
@@ -0,0 +1,32 @@
+-- |
+-- Module      :  Pact.Core.IR.Typecheck
```
rsoeldner, line 2:
```suggestion
-- Module      :  Pact.Core.Version
```

rsoeldner (approved):

PR #179, opened 2024-07-15
Support transaction batch signatures

imalsogreg:
This is a very WIP, unstable PR, because we need a KIP for a batch signature protocol before we merge anything into the pact server.

PR checklist:

* [ ] Benchmark regressions


pact-lsp/Pact/Core/LanguageServer.hs
```diff
@@ -232,7 +232,7 @@ setupAndProcessFile nuri content = do
   gasLog <- newIORef Nothing
   let
     builtinMap = if isReplScript fp
-                 then replCoreBuiltinMap
+                 then replBuiltinMap -- TODO: Correct?
```
rsoeldner, line 235: 👍

PR #191, opened 2024-07-29
Add Necessary encodings for Local, Add ExecutionMode to the Evaluate API

jmcardon:
Changes in this PR are purely for chainweb's API and thus do not impact pact's test suite.

Test Coverage is done on chainweb directly.

PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact-request-api/Pact/Core/Command/Types.hs
+++ b/pact-request-api/Pact/Core/Command/Types.hs
@@ -85,7 +85,6 @@ import Pact.Core.Gas.Types
 import Pact.Core.Names
 import qualified Pact.Core.Hash as PactHash
 import Pact.Core.Persistence.Types
-import Pact.Core.Info
 import Pact.Core.PactValue (PactValue(..))
 import Pact.Core.Command.RPC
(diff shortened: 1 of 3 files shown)


pact-request-api/Pact/Core/Command/Types.hs
```diff
@@ -145,16 +145,13 @@ type SigCapability = CapToken QualifiedName PactValue
 -- | Pair parsed Pact expressions with the original text.
 data ParsedCode = ParsedCode
   { _pcCode :: !Text
-  , _pcExps :: ![Lisp.TopLevel ()]
+  , _pcExps :: ![Lisp.TopLevel Info]
```
rsoeldner, line 148: Why this change to `SpanInfo`?

jmcardon: Good question. 

We want chainweb's choice of `Info` to be governed by the `Info` alias in `Evaluate.hs`

`ParsedCode` was nuking line information before. Now we get to keep it and be consistent.

rsoeldner (approved):

PR #206, opened 2024-08-23
Stop charging for hashes and SpanInfos

edmundnoble:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/SizeOf.hs
+++ b/pact/Pact/Core/SizeOf.hs
@@ -12,6 +12,7 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE DerivingVia #-}
 {-# LANGUAGE DataKinds #-}
+{-# LANGUAGE NamedFieldPuns #-}
 
 -- |
 -- Module      :  Pact.Types.SizeOf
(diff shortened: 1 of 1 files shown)


pact/Pact/Core/SizeOf.hs
```diff
@@ -436,13 +438,44 @@ instance (SizeOf t, SizeOf i) => SizeOf (DefSchema t i)
...
+      , _mGovernance
+      , _mDefs
+      , _mBlessed
+      , _mImports
+      , _mImplements
+      , _mHash
+      , _mTxHash
+      , _mInfo
+      } -> sum <$> sequence
+        [ sizeOf i ver _mName
```
jmcardon, line 454: I'm not particularly sure how to feel about these changes tbh. The idea of the gas charge of infos is meant to actually charge on the memory occupied by the module. Moreover, it doesn't exactly change that `sizeOf i ver _mDefs` is still charging for infos.

pact/Pact/Core/SizeOf.hs
```diff
@@ -381,7 +382,8 @@ instance SizeOf Yield
 instance SizeOf DefPactExec
 
 -- spaninfo
-instance SizeOf SpanInfo
+instance SizeOf SpanInfo where
+  sizeOf _ _ _ = pure 0
```
jmcardon, line 386: Charging 0 for a structure that accounts for something like ~30% of the actual module size on disk seems wrong to me.

jmcardon (commented):
So with this review, there's something to point out: `SizeOf` is not a great abstraction, and needs to be removed/replaced within pact. Ideally before pact-5 release, but afterwards is fine as well.

As an abstraction, it measures how large a Haskell structure is in memory. This might be particularly useful for an allocator or for anything that keeps track of allocation, but the actual current use of it is to charge for writes/reads to/from the database. That makes `sizeOf`'s current use case not particularly great, because the size of something in particular as it is written to disk depends more on the encoding. 

That said: if decoding line information takes up a reasonable chunk of time when reading off of disk, or simply takes up a reasonable chunk of memory, it does not seem good to zero out its cost. That can mean gas doesn't capture the actual behavior intended.

PR #208, opened 2024-08-29
Add gas to post-reads, make module admin acquisition explicit.

jmcardon:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/IR/Eval/CEK.hs
+++ b/pact/Pact/Core/IR/Eval/CEK.hs
@@ -18,7 +18,6 @@ module Pact.Core.IR.Eval.CEK
   , resumePact
   , evalCap
   , nameToFQN
-  , guardTable
   , isKeysetInSigs
   , isKeysetNameInSigs
   , requireCap
@@ -157,13 +156,11 @@ evaluateTerm cont handler env (Var n info) = do
           failInvariant info (InvariantInvalidDefKind (defKind mname d) "in var position")
         Nothing ->
           failInvariant info (InvariantInvalidBoundVariable (_nName n))
-    NModRef m ifs -> case ifs of
-      [] -> throwExecutionError info (ModRefImplementsNoInterfaces m)
-      _ ->
+    NModRef m ifs ->
         returnCEKValue cont handler (VModRef (ModRef m (S.fromList ifs)))
     NDynRef (DynamicRef dArg i) -> case RAList.lookup (view ceLocal env) i of
       Just (VModRef mr) -> do
-        modRefHash <- _mHash <$> getModule info (view cePactDb env) (_mrModule mr)
+        modRefHash <- _mHash <$> getModule info (_mrModule mr)
         let nk = NTopLevel (_mrModule mr) modRefHash
         evalCEK cont handler env (Var (Name dArg nk) info)
       Just _ ->
@@ -363,7 +360,7 @@ applyPact
   -> EvalM e b i (EvalResult e b i)
 applyPact i pc ps cont handler cenv nested = use esDefPactExec >>= \case
   Just pe -> throwExecutionError i (MultipleOrNestedDefPactExecFound pe)
-  Nothing -> getModuleMemberWithHash i (_cePactDb cenv) (pc ^. pcName) >>= \case
+  Nothing -> getModuleMemberWithHash i (pc ^. pcName) >>= \case
     (DPact defPact, mh) -> do
       let nSteps = NE.length (_dpSteps defPact)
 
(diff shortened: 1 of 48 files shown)


pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -1170,45 +1141,45 @@ applyContToValue (BuiltinC env info frame cont) handler cv = do
...
-          evalWrite info pdb wt (tvToDomain tv) rk rdata
-          returnCEKValue cont handler (VString "Write succeeded")
-        else
-          throwExecutionError info (WriteValueDidNotMatchSchema (_tvSchema tv) (ObjectData rv))
-      PreFoldDbC tv queryClo appClo -> do
-        let tblDomain = DUserTables (_tvName tv)
-        -- Todo: keys gas
-        keys <- liftDbFunction info (_pdbKeys pdb tblDomain)
-        foldDBRead tv queryClo appClo keys []
+      -- ReadC tv rowkey -> do
```
rsoeldner, line 1152: cleanup

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -1221,17 +1192,17 @@ applyContToValue (BuiltinC env info frame cont) handler cv = do
...
-        returnCEKValue cont handler (VString "TableCreated")
-      EmitEventC ct@(CapToken fqn _) -> do
-        d <- getDefCap info fqn
-        enforceMeta (_dcapMeta d)
-        emitCapability info ct
-        returnCEKValue cont handler (VBool True)
-        where
-        enforceMeta Unmanaged = throwExecutionError info (InvalidEventCap fqn)
-        enforceMeta _ = pure ()
+      -- CreateTableC (TableValue tn _ _) -> do
```
rsoeldner, line 1195: same

pact/Pact/Core/IR/Eval/CEK.hs
```diff
@@ -1703,13 +1674,16 @@ eval purity benv term = do
 evalWithinCap
   :: forall e b i
   .  ()
-  => Purity
+  => i
+  -> Purity
   -> BuiltinEnv e b i
-  -> CapToken FullyQualifiedName PactValue
+  -> CapToken QualifiedName PactValue
```
rsoeldner, line 1680: Why are we moving from FQN to QN, especially as we need fqn (see line 1686).?

pact/Pact/Core/IR/Eval/CEK/Types.hs
```diff
@@ -380,35 +379,22 @@ data BuiltinCont (e :: RuntimeMode) (b :: K.Type) (i :: K.Type)
   -- ^ {closure} {accum} {rest}
   | ZipC (CanApply e b i) ([PactValue],[PactValue]) [PactValue]
   -- ^ <zip closure> <lists to zip> <accumulator>
-  | PreSelectC TableValue (CanApply e b i) (Maybe [Field])
-  -- ^ <table> <select filter closure> <filter fields>*
-  | PreFoldDbC TableValue (CanApply e b i) (CanApply e b i)
-  -- ^ <table> <select filter closure> <accumulator closure>
+  -- | PreSelectC TableValue (CanApply e b i) (Maybe [Field])
```
rsoeldner, line 382: cleanup

pact/Pact/Core/IR/Eval/CEK/Utils.hs
```diff
@@ -94,7 +94,7 @@ sysOnlyEnv e
   let newPactdb =
           PactDb
          { _pdbPurity = PSysOnly
-         , _pdbRead = read'
+         , _pdbRead = \dom k -> read' dom k
```
rsoeldner, line 97: This change seems unnecessary

pact/Pact/Core/IR/Eval/Direct/Evaluator.hs
```diff
@@ -153,6 +154,26 @@ eval purity benv term = do
...
+  => i
+  -> Purity
+  -> BuiltinEnv e b i
+  -> CapToken QualifiedName PactValue
+  -> EvalTerm b i
+  -> EvalM e b i PactValue
+evalWithinCap info purity benv (CapToken qualName vs) term = do
+  ee <- viewEvalEnv id
+  (_, mh) <- getDefCapQN info qualName
+  let ct = CapToken (qualNameToFqn qualName mh) vs
```
rsoeldner, line 170: Maybe we should go with only FQN?

rsoeldner (approved):

pact/Pact/Core/IR/Eval/CEK.hs
jmcardon: Great question, it's because this is particularly so the API is easier to use. This is generally called by chainweb to call something like evaluating something within`(some_module.GAS_PAYER args)`. We don't want our caller to have to deal with getting the fully qual name themselves

pact/Pact/Core/IR/Eval/Direct/Evaluator.hs
jmcardon: See: Other comment about this same change. This is for external facing apis

PR #210, opened 2024-08-30
Fix captoken codec, add SigCapability, move Signer to its own module

jmcardon:
This PR moves `Signer` into its own module, as it isn't really used outside of the chainweb and request API.

Moreover: I am adding a `SigCapability` newtype, that essentially has a fixed json codec for the sake of both chainweb, as well as testing.

The changes here are mostly a refactoring, so it's covered by previous tests for signers. That said: in fixing the codec of `ModuleGuard`, I realize we definitely need a golden test for stable encoding. This will be done by @rsoeldner 

PR checklist:

* [x] Benchmark regressions
Doesn't impact perf


rsoeldner:
We approved in a call, forgot to approve it on GH.

PR #214, opened 2024-09-04
LSP: Simplify parser and parse tree for LSP

jmcardon:
This PR is a prelude that's necessary for further LSP work. It essentially simplifies the parser and moves a lot of the work that was previously done in the tokenizer + parser into the desugarer.

The parse tree is thus simplified significantly and far more independent of what we need for execution, which will help our LSP provide more accurate information on hover (the LSP analyzer is coming in a subsequent PR)

It also improves repl docs. That is, pre this pr:

```
pact>with-capability
(interactive):1:0: Unexpected input after expr, remaining tokens: [with-capability]
 1 | with-capability
   | ^^^^^^^^^^^^^^^
```

Post:
```
pact>with-capability
## with-capability

Use `with-capability` to apply the access to a specific capability to execute a body of code.
This function ensures that an elevated privilege—defined as a capability using a `defcap` code block—is present during the execution of the provided body of code. 

You can only call the `with-capability` function in the same module that contains the corresponding `defcap` declaration. 
If the token that grants permission to use the specified capability isn't found, the `with-capability`  evaluates the capability definition to install or grant the permission token. 
The permission token is automatically revoked after executing the code body. 
Nested `with-capability` calls for the same permission token detect the presence of the token and execute the body without reapplying the capability.

...
```

PR checklist:

* [x] Benchmark regressions
  - Benchmarks show no difference in execution speed


pact-tests/gas-goldens/builtinGas.golden
```diff
@@ -81,7 +81,7 @@ not: 139
 not?: 139
 or?: 139
 pact-id: 60000271
-pairing-check: 67003096
+pairing-check: 67003097
```
rsoeldner, line 84: I'm wondering what has caused the pairing-check gas increase?

pact-tests/pact-tests/yield.repl
```diff
@@ -92,7 +92,7 @@
 
 ;; check events
 (expect "step 0 events"
- [{"module-hash": "oKFbzkuEYAFhP-S2mW7hRvRJUPJf2FvMFy1CpxhGs4o"
+ [{"module-hash": "rmN99MpBmJbapgVRV3GjII6I4UUkl8k5pBF7-k92Lt8"
```
rsoeldner, line 95: isn't the module-hash based on stable encoding? It shouldn't change?

pact/Pact/Core/Syntax/LexUtils.hs
```diff
@@ -80,26 +79,18 @@ data Token
...
-  | TokenEnforceOne
   | TokenSingleTick !Text
   | TokenIdent !Text
   | TokenNumber !Text
   | TokenString !Text
   | TokenTrue
   | TokenFalse
-  | TokenBlockIntro
-  | TokenSuspend
+  -- | TokenBlockIntro
```
rsoeldner, line 88: Cleanup

pact/Pact/Core/Syntax/ParseTree.hs
```diff
@@ -377,32 +367,29 @@ data CapForm i
...
+  | Lam [MArg i] (NonEmpty (Expr i)) i
   | App (Expr i) [Expr i] i
-  | Block (NonEmpty (Expr i)) i
-  | Operator Operator i
   | List [Expr i] i
   | Constant Literal i
-  | Try (Expr i) (Expr i) i
-  | Suspend (Expr i) i
   | Object [(Field, Expr i)] i
+  -- | Block (NonEmpty (Expr i)) i
```
rsoeldner, line 389: same

rsoeldner (approved):

pact-tests/pact-tests/yield.repl
jmcardon: It's not based on stable encoding. It _should_ change. It's based on our CBOR encoding.

PR #217, opened 2024-09-06
Fix verifier codec

jmcardon:
#210 unintentionally broke a codec chainweb depends on. This PR fixes the signer codec + adds roundtrip tests

PR checklist:

* [ ] Benchmark regressions


jmcardon:
Note: This PR is blocked on making CI pass. We need to ensure that we have the regression sqlite file cached in the runners somehow

rsoeldner (approved):

PR #228, opened 2024-09-13
Integration: Add legacy pact id support

jmcardon:
This is to support previously clobbered pact ids, where the `show` instance was used for the key. This bleeds over to chainweb.

PR checklist:

* [x] Benchmark regressions


jmcardon:
Closing, we want this in chainweb instead

PR #237, opened 2024-09-21
fix tarball file permissions to allow  write access for tarball extractor

salamaashoush:
It resolves the writable permissions issue for the tarball and removes the Nix path from the bundled files, creating a flat directory structure like `/bin` instead of `/nix/store/<nix-stuff>/bin`. This approach follows the conventional method for distributing tarballs.


DevopsGoth (approved):
names changed in main release to match standards

PR #249, opened 2024-10-09
Add MPFR trans functions

rsoeldner:
Based on https://github.com/kadena-io/pact/pull/1045/files#diff-9619507e39a18e5493fafa72986db085aebadb61dc76aac3396fc34e30d38623

PR checklist:

* [ ] Benchmark regressions


jmcardon:
I actually want to get #250  in before this one.

Essentially: I am still tweaking gas a bit, and that PR allows for gassing `pow` in an unbounded fashion. As is, we can't merge MPFR without it I think

jmcardon (approved):

PR #253, opened 2024-10-10
DB error messages

lsgunnlsgunn:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/Errors.hs
+++ b/pact/Pact/Core/Errors.hs
@@ -1225,32 +1225,32 @@ ensureBound msg = BoundedText (T.take (fromIntegral (natVal (Proxy @k))) msg)
 dbOpErrorToBoundedText' :: DbOpError -> Text
 dbOpErrorToBoundedText' = \case
     WriteError ->
-      "Error found while writing value"
+      "Database error: Writing " rk "in" tn " failed because of an error."
     RowReadDecodeFailure rk ->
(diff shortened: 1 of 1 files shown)


pact/Pact/Core/Errors.hs
```diff
@@ -1225,32 +1225,32 @@ ensureBound msg = BoundedText (T.take (fromIntegral (natVal (Proxy @k))) msg)
...
+        ["Update failed because no row was found in table:"
         , renderTableName tn
         , "at key"
         , tdquotes $ abbrevRowKey rk ]
     NoSuchTable tn ->
-      thsep ["Table", renderTableName tn, "not found"]
+      thsep ["Insert/Update failed because table ", renderTableName tn, "was not found."]
     TableAlreadyExists tn ->
-      thsep ["Table", renderTableName tn, "already exists"]
+      thsep [""Insert/Update failed because table", renderTableName tn, "already exists.""]
```
rsoeldner:
```suggestion
      thsep [""Creating of user table", renderTableName tn, "failed because it already exists."]
```

lsgunnlsgunn: I can't commit the suggestion but with a minor tweak this looks good to me (I'm not seeing the error messages in context, but if table creation fails in this case, the message looks right to me).

PR #273, opened 2024-10-31
Run-read-only native

jmcardon:
cc @CryptoPascal31

[Implements KIP 0024](https://github.com/kadena-io/KIPs/pull/53). 

The main difference is in the name. I felt that `pure` isn't really clear to users what it is, but `run-read-only` does. Moreover, `pure` has a different connotation in programming (a function that has no side effects), and also internally, the mode is called `read-only` in pact. 

I'm still open to suggestions on the naming

PR checklist:

* [x] Benchmark regressions

### DIFF
--- a/pact-lsp/Pact/Core/LanguageServer/Utils.hs
+++ b/pact-lsp/Pact/Core/LanguageServer/Utils.hs
@@ -34,6 +34,7 @@ termAt p term
           CWithCapability a b -> termAt p a <|> termAt p b
           CTry a b -> termAt p a <|> termAt p b
           CCreateUserGuard a -> termAt p a
+          CRunReadOnly a -> termAt p a
         <|> Just t
       t@(ListLit tms _) -> getAlt (foldMap (Alt . termAt p) tms) <|> Just t
       t@(Nullary tm _) -> termAt p tm <|> Just t
--- a/pact-lsp/Pact/Core/LanguageServer/Utils.hs
+++ b/pact-lsp/Pact/Core/LanguageServer/Utils.hs
@@ -34,7 +34,7 @@ termAt p term
           CWithCapability a b -> termAt p a <|> termAt p b
           CTry a b -> termAt p a <|> termAt p b
           CCreateUserGuard a -> termAt p a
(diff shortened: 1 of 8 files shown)


emilypi (approved):

CryptoPascal31:
Thank you for implementing it... This will be a small but a significant step for improving the Pact security model.

The advantage of `pure` is that this is very concise: 4 letters. This is subjective, but IMHO it's important for "utility" natives.
And AFAIK in Pact, pure is synonym to read-only, as only DB writes can change the state of the Pact interpreter.

Maybe a tradeoff between concision and self-explanatory should be `(read-only    )` instead of `(run-read-only   )`

jmcardon:
> The advantage of pure is that this is very concise: 4 letters. This is subjective, but IMHO it's important for "utility" natives.

This, I don't necessarily think is true. I do agree that this is definitely very subjective, and I don't necessarily think that the native needs to be short.

> And AFAIK in Pact, pure is synonym to read-only, as only DB writes can change the state of the Pact interpreter.

This is true. It could be my personal bias here. `pure` to me means something different (no side effects). I think I might be a bit nitpicky and `pure` is likely fine.

> Maybe a tradeoff between concision and self-explanatory should be (read-only    ) instead of (run-read-only   )

I frankly prefer `pure` to `read-only`. `run-read-only` made sense to me because the expression `e` passed in to `(run-read-only e)` is evaluated in read-only mode, and thus `run`. I also envisioned as reading code that does this as:
```
(run-read-only 
  (do
    (thing)
    (thing2)
    (thing3))
 )
 ```
 
 I think this reads pretty cool, it basically says "run this whole block of expressions in read only mode". It's almost self explanatory. I think the equivalent 
 ```
(pure
  (do
    (thing)
    (thing2)
    (thing3))
 )
 ```
 
 Is likely fine as well (as much as it may confuse haskellers such as myself haha), it just didn't have the same ring. I am definitely amenable to renaming it and discussing it here or in the KIP, but preferably here.

rsoeldner (approved):

PR #274, opened 2024-10-31
Enforce and user guards as read-only

jmcardon:
cc @CryptoPascal31

[Implements KIP 0027](https://github.com/kadena-io/KIPs/pull/56), and loosens the restriction that `enforce` should be sys-only as well. This was probably the for FV's sake, but considering we need to revamp FV as well, I think it makes sense to loosen this restriction.

It also adds regression tests for this behaviour as well.

As a bit of historical background note: pact did not used to always have runtime recursion checking, but now that we do, I think it's probably fair to consider this as a straight enhancement. We are still not vulnerable to any sort of reentrant attacks due to the mode being set to read-only.

PR checklist:

* [x] Benchmark regressions


CryptoPascal31:
Ty

rsoeldner (approved):

PR #277, opened 2024-11-04
Chore: Separate repl into separate project, fix REPL docs output

jmcardon:
This PR separates the REPL into its own sub-library which helpful for fixing several issues:
- This improves the compile times of pact-5 for the sake of chainweb integration.
- It allows us to pull in `pandoc`, which fixes in our docs issues where our docs in the repl look like the raw markdown docs.

Before this PR: 
image

After this PR:
image

PR checklist:

* [x] Benchmark regressions
- N/A. 


jmcardon:
As of my last commit, it also fixes an annoying issue in the pact-5 repl that repl functions were actually going through the gas charging method, which when it actually charges gas, will essentially brick the repl. 

The following is now possible:
```
pact>(env-gasmodel "table")
"Set gas model to table-based cost model"
pact>(env-gaslimit 1)
"Set gas limit to 1"
pact>(env-gas
env-gas             env-gaslimit        env-gaslog          env-gasmodel        env-gasmodel-fixed
pact>(+ "hello" "world")
(interactive):1:0: Gas limit "1"exceeded: 2
 1 | (+ "hello" "world")
   | ^^^^^^^^^^^^^^^^^^^


pact>(env-gas 0)
"Set gas to 0"
```

docs/builtins/General/enforce.md
```diff
@@ -18,7 +18,7 @@ Use the following arguments to specify the test expression and error message for
 
 | Argument | Type   | Description                                    |
 |----------|--------|------------------------------------------------|
-| `expression` | bool | Specifies the expression to evaluate.     |
+| expression | bool | Specifies the expression to evaluate.     |
```
rsoeldner, line 21: @lsgunnlsgunn last PR's added the backticks. What is the _correct_ style now, see: https://github.com/kadena-io/pact-5/pull/278/files

rsoeldner (approved):

docs/builtins/General/enforce.md
jmcardon: I don't actually understands why this breaks docusaurus, but it looks weird in our ansi terminal renderer, but lisa can add them back if necessary

lsgunnlsgunn: We don't need backticks around ordinary words or words to indicate syntax, per se. We only *require* them for certain characters that freak Docusaurus out where the docs engine doesn't know how to interpret things like < and > or { and } except as HTML. I don't know what all the characters are that it doesn't like. Amir put something together that records them, I think.

PR #287, opened 2024-12-03
Fix issues with runtime typechecker

jmcardon:
The pact 5 runtime TC was skipping let binds, as well as not mangling some interface names correctly which resulted in typechecking failure false positives.

PR checklist:

* [x] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/IR/Desugar.hs
+++ b/pact/Pact/Core/IR/Desugar.hs
@@ -1389,7 +1389,7 @@ renameModule
   :: (DesugarBuiltin b)
   => Module ParsedName DesugarType b i
   -> RenamerM e b i (Module Name Type b i)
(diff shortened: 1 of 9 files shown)


pact/Pact/Core/IR/Desugar.hs
```diff
@@ -1461,8 +1462,9 @@ checkImplements
   -> RenamerM e b i ()
 checkImplements i defs moduleName ifaceName = do
   resolveModuleData ifaceName i >>= \case
-    InterfaceData iface _deps ->
+    InterfaceData iface _deps -> do
       traverse_ checkImplementedMember (_ifDefns iface)
+      pure ()
```
edmundnoble, line 1467: Not needed

edmundnoble (approved):

chessai (approved):

PR #305, opened 2025-01-23
Github Actions @v3 -> @v4

DevopsGoth:
PR checklist:

* [ ] Benchmark regressions


DevopsGoth:
There's actually a ton of version updates to the workflows that will be more future-proof.  Merging when these builds run successfully.

PR #318, opened 2025-02-07
update links with new pact docs urls

CarlParkerWeb3:
PR checklist:

* [ ] Benchmark regressions


CarlParkerWeb3:
Was on my radar before yours, didn't think to check again when sending. Considering the 2 links in Docs section. 2nd link we both have the same. 1st link, I would the one I suggested, not the generic landing page.  

@rsoeldner

rsoeldner:
thank you @CarlParkerWeb3, highly appreciate your support. Lisa continued this in #319

PR #328, opened 2025-02-12
Add checks for native shadowing in REPL and shadowing check module

jmcardon:
In pact 5, we disallow natives to be shadowed. if you have an expression such as
```
(let* ((identity 1)) identity)
```
`identity` is actually resolved as the `identity` native `(lambda (x) x)`. This is actually incredibly annoying to debug because pact 4 allowed this, but we do not, since the error received is usually "Expected pact value, got closure" whenever you return a variable that collides with a native name.

This PR turns this into an error in the REPL, and we have plans to add this to mainnet on our next service release. Past a certain fork height, transactions where native names shadow variables will not make it into the mempool.

PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact-repl/Main.hs
+++ b/pact-repl/Main.hs
@@ -52,6 +52,7 @@ data ReplOpts
   | OServer FilePath
   -- Crypto
   | OGenKey
+  | OCheckNativeShadowing FilePath
   deriving (Eq, Show)
 
 replOpts :: O.Parser (Maybe ReplOpts)
@@ -61,8 +62,10 @@ replOpts = O.optional $
   <|> O.flag' OLanguageServer (O.long "lsp" <> O.help "Start Language Server")
   <|> apiReqFlag
   <|> unsignedReqFlag
+  <|> O.flag' OGenKey (O.short 'g' <> O.long "genkey" <> O.help "Generate ED25519 keypair")
   <|> loadFlag
   <|> OServer <$> O.strOption (O.metavar "CONFIG" <> O.short 's' <> O.long "server" <> O.help "Run Pact-Server")
+  <|> checkNativeShadowingFlag
 
 -- Todo: trace output and coverage?
 loadFlag :: O.Parser ReplOpts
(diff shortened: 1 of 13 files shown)


pact-repl/Main.hs
```diff
@@ -76,6 +79,11 @@ loadFlag = fmap OLoad $
     <*> O.argument O.str
       (O.metavar "FILE" <> O.help "File path to compile (if .pact extension) or execute.")
 
+checkNativeShadowingFlag :: O.Parser ReplOpts
```
rsoeldner, line 84: Shouldn't this be a flag (no filename) ?

pact/Pact/Core/Errors.hs
```diff
@@ -1763,6 +1767,8 @@ desugarErrorToBoundedText = mkBoundedText . \case
       thsep ["Duplicate definition:", renderQualName qn]
     InvalidBlessedHash hs ->
       thsep ["Invalid blessed hash, incorrect format:", hs]
+    InvalidNativeShadowing t ->
+      thsep ["Variable", t, "shadows native with same name"]
```
rsoeldner, line 1771:
```suggestion
      thsep ["Variable", t, "shadows native with the same name"]
```

pact/Pact/Core/NativeShadowing.hs
```diff
@@ -0,0 +1,255 @@
...
+import Pact.Core.Type
+import Pact.Core.Pretty
+
+import qualified Data.Map.Strict as M
+
+-- | The context of what is being shadowed,
+--   that is, the location and type of definition where shadowing is occuring.
+data ShadowCtx
+  = DefNameCtx DefKind QualifiedName
+  | DefArgContext DefKind Int QualifiedName
```
rsoeldner, line 42: Can you add some comments here

rsoeldner (approved):

pact-repl/Main.hs
jmcardon: This one isn't a flag for the repl, this allows you to check shadowing on a file without actually running it. 

In the repl, checking native shadowing will be enabled unconditionally.

PR #329, opened 2025-02-12
Remove ubuntu 20.04 builds

DevopsGoth:
Removing Ubuntu 20.04, adding 24.04 (which is current ubuntu-latest).

Ubuntu 20.04 runners are being sunset next month.

PR checklist:

* [ ] Benchmark regressions


rsoeldner (approved):
This seems to be isync with chainweb-node: https://github.com/kadena-io/chainweb-node/blob/master/.github/workflows/applications.yml (at least ubuntu 22.04)

Thank you 👍

PR #331, opened 2025-02-14
Calculate minimal dependency set on module deploys

jmcardon:
PR checklist:

* [x] Benchmark regressions


pact-tests/pact-tests/db.repl
```diff
@@ -150,26 +150,26 @@
...
-    (insert t "b" { 'x: 2 })
-    (rollback-tx)
-    (expect "2nd insert rolled back" ["a"]
-            (keys t))))
-
-(create-table t)
-(commit-tx)
-
-(nested-tx.test-nested-tx)
+;  (begin-tx)
```
jmcardon, line 153: This test needs to be rewritten, but basically, `defun test-nested-tx` is an ugly corner case of the language we allow in the repl, but now we cannot allow it to preserve free variables across `begin-tx` boundaries. This test needs minor rewriting.

rsoeldner (approved):

PR #335, opened 2025-02-24
Fix local in pact server

### LINKED ISSUE #326: Local endpoint executes write transactions in local environment
## Issue description

In the previous versions transactions to the "/local" endpoint in local environment worked similarly to the testnet and mainnet environments, that is they did not execute write transactions. After upgrading my dev environment to Pact 5 the local call actually inserts data into the SQLite database and therefore you get an error if your code tests the transaction against the local endpoint and then sends the same command to the "/send" endpoint. Is this change in behaviour intended?

### Steps to reproduce

1. Install Pact 5 on your local environment
2. Send a transaction which inserts data to blockchain to the local endpoint (for example create a "coin" account):
pact -a call.yaml | curl -H "Content-Type: application/json" -d @- http://localhost:8080/api/v1/local
3. Send the same transaction to the send endpoint:
pact -a call.yaml | curl -H "Content-Type: application/json" -d @- http://localhost:8080/api/v1/send

### Expected Behavior

Both calls should succeed and only the second call should actually insert a row to the database.

jmcardon:
Closes #326 

PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact-request-api/Pact/Core/Command/Server.hs
+++ b/pact-request-api/Pact/Core/Command/Server.hs
@@ -277,7 +277,7 @@ sendHandler runtime (SendRequest submitBatch) = do
     requestKeys <- forM (_sbCmds submitBatch) $ \cmd -> do
       let requestKey = cmdToRequestKey cmd
       res <- liftIO $ try $! do
-        result <- computeResultAndUpdateState runtime requestKey cmd
+        result <- computeResultAndUpdateState runtime Transactional requestKey cmd
         storeResult <- newEmptyMVar
         writeChan (_srvChan runtime) (StoreMsg requestKey result storeResult)
         readMVar storeResult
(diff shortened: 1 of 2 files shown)


mvaljento:
@edmundnoble Any chance of getting this PR reviewed? This local endpoint issue is messing with my project. :)

mvaljento:
@edmundnoble Ah, I see. Sorry to harass you! I shall wait.

mvaljento:
I installed the latest nightly build, which should include this fix and indeed the local endpoint does not write a row to database anymore, it still throws the following error when submitting the same transaction first to /local endpoint and then to the /send endpoint with the identical nonce.

"Request key: zmOagLSmC7X-wvHQXcBaPP-Eoqb0ohIiAWKpzckB0Yk has already been submitted. If you are resubmitting a valid transaction, try changing the nonce in the payload"

Is this a feature or a bug? As far as I understand the send endpoint should accept the same nonce as the local endpoint as the local endpoint transaction is not written to the blockchain? Or is it by design that the caller should change the nonce for every transaction? 

It used to work with the same nonce. If it's an intended change it requires refactoring all code that tests the identical command against the local endpoint and should be pointed out in docs if not already done so.

jmcardon:
> @jmcardon
> 
> I installed the latest nightly build, which should include this fix and indeed the local endpoint does not write a row to database anymore, but it still throws the following error when submitting the same transaction first to /local endpoint and then to the /send endpoint with the identical nonce.
> 
> `"Request key: zmOagLSmC7X-wvHQXcBaPP-Eoqb0ohIiAWKpzckB0Yk has already been submitted. If you are resubmitting a valid transaction, try changing the nonce in the payload"`
> 
> As far as I understand this is still a bug, since using the same nonce works in testnet (the transaction is not being "resubmitted").
> 
> I will open a new issues, if no objections.

Yes please do 🙏 . 

I'm surprised that this is the case. I will look into it.

PR #336, opened 2025-02-25
Flag function as no longer supported

lsgunnlsgunn:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/docs/builtins/Database/txlog.md
+++ b/docs/builtins/Database/txlog.md
@@ -1,6 +1,7 @@
-## txlog
+## txlog (DEPRECATED)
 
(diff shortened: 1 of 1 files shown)


jmcardon:
Why don't we just delete the file entirely?

jmcardon (approved):

lsgunnlsgunn:
> Why don't we just delete the file entirely?

I have no problem with that solution, but it looks to me like we haven't done that in the past (env-keys, decrypt-cc20p1305, and validate-keypair, for example), so I thought there might be some kind of legacy reason for keeping old functions around

PR #338, opened 2025-02-25
Add gas for transitive dependencies

jmcardon:
This adds gas metering as well as benchmarks for our transitive dependency pass.

The following graph comes from the benchmarks, run on MacOS Sonoma with an M2 MAX chip:
```
➜  pact-core git:(jose/bench-transdeps) sysctl -a | grep machdep.cpu
machdep.cpu.cores_per_package: 12
machdep.cpu.core_count: 12
machdep.cpu.logical_per_package: 12
machdep.cpu.thread_count: 12
machdep.cpu.brand_string: Apple M2 Max
```
image

The X axis gas charged (Actual), and Y is time spent, converted to gas units (1 milligas / 1 ns). We do not want these to be equivalent, only proportional, due to variances in execution based on hardware/etc.

PR checklist:

* [X] Benchmark regressions


gasmodel/Pact/Core/GasModel/ModuleDepsBench.hs
```diff
@@ -0,0 +1,328 @@
...
+type DepGenM a = ReaderT (ModuleName, ModuleHash) (StateT DPGenState Gen) a
+
+
+emptyModule :: ModuleName
+emptyModule = ModuleName "" Nothing
+emptyModuleHash :: ModuleHash
+emptyModuleHash = ModuleHash (Hash "")
+
+-- | Run a DepGenM action and return the set of all generated definitions
+runDepGenM :: DepGenM a -> Gen (a, M.Map FullyQualifiedName (EvalDef CoreBuiltin LineInfo))
```
jmcardon, line 85: Most of our `DepGenM` code is copy pasted from `TransitiveDependencyTests.hs` but with minor tweaks to intentionally generate larger modules for benchmarks.

pact-tests/Pact/Core/Test/ReplTests.hs
```diff
@@ -89,7 +89,7 @@ runReplTest (ReplSourceDir path) pdb file src interp = do
       let rendered = renderLocatedPactErrorFromState rstate' e
       assertFailure (T.unpack rendered)
     Right _ -> do
-      traverse_ ensurePassing . _replTestResults =<< readIORef stateRef
+      traverse_ ensurePassing . reverse . _replTestResults =<< readIORef stateRef
```
jmcardon, line 92: Btw this reverse is just a QOL fix.

I noticed when debugging values by running `cabal run tests -- -p 'coin-v5'` that the errors being spit out by the repl tests were in reverse order.

pact/Pact/Core/TransitiveDependencies.hs
```diff
@@ -111,36 +170,85 @@ canHaveDependents = \case
...
+
+-- | Compute all transitive dependencies from our initial working set.
+--   The algorithm is pretty simple: If a particular component can have dependents,
+--   then we traverse its tree and collect all of its dependents, and push those not currently in
+--   the dependency set into it as well as the working set.
+--
+--   The algorithm only ever grows the dependency set, it will never shrink it, thus
+--   it is guaranteed to terminate
+computeTransitiveDeps :: TransitiveClosureM b i ()
+computeTransitiveDeps = computeUntilDone
```
jmcardon, line 194: This is my favorite part of this PR. By floating the algorithm out of `getAllTransitiveDependencies`, it simplified _drastically_

pact/Pact/Core/TransitiveDependencies.hs
```diff
@@ -49,37 +57,88 @@ addToWorkingSet fqn (WorkingSet s) = WorkingSet (S.insert fqn s)
 popFromWorkingSet :: WorkingSet -> Maybe (FullyQualifiedName, WorkingSet)
 popFromWorkingSet (WorkingSet s) = over _2 WorkingSet <$> S.minView s
 
-type TransitiveClosureM a = State TransitiveClosureState a
+type TransitiveClosureM b i a = ExceptT TransitiveClosureCalcErr (State (TransitiveClosureState b i)) a
+
+costPerFQNComparison :: SatWord
```
jmcardon, line 62: This should be in `GasCostConfig`, but this is minor and can be done in a followup, and would not be forking.

pact/Pact/Core/TransitiveDependencies.hs
```diff
@@ -111,36 +170,85 @@ canHaveDependents = \case
...
+                    , _gasConsumed = currGasConsumed
+                    , _gasLimit = mgasLimit
+                    , _allDependencies = allDeps }
+    case runState (runExceptT computeTransitiveDeps) tcloState of
+      (c, TransitiveClosureState _ws (DependencySet depSet) (MilliGas consumedByAlgorithm) _glim _allDeps) -> do
+        -- The way this module works, it starts at the current gas consumed, but does not charge gas
+        -- to our gas ref directly, so what we want is the difference of gas consumed post algorithm
+        -- exec vs gas pre-algorithm exec
+        -- We charge gas regardless here, which may raise
+        let consumedAmt = consumedByAlgorithm - currGas + restrictKeysCost allDeps depSet
```
edmundnoble, line 240: This is funny. We do almost the exact same thing for verifiers. Perhaps it should be a library function?

jmcardon: Possibly, we might want a general interface for this?

PR #346, opened 2025-02-27
Fix ordering of writes broken by transitive deps PR

jmcardon:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact/Pact/Core/Compile.hs
+++ b/pact/Pact/Core/Compile.hs
@@ -211,7 +211,6 @@ evalTopLevel interpreter (RawCode code) tlFinal deps = do
       -- can't fix it later.
         CapGov _ -> pure ()
       -- Write sliced modules to the pact db
-      evalWrite (_mInfo m) pdb Write DModuleSource (getHashedModuleName m) (ModuleCode code)
 
       -- Get all of the _new_ dependencies (that is, members of the deployed module)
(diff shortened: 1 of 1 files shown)


pact/Pact/Core/Compile.hs
```diff
@@ -229,11 +228,13 @@ evalTopLevel interpreter (RawCode code) tlFinal deps = do
...
       chargeGasArgs (_mInfo m) (GWrite mSize)
       evalWrite (_mInfo m) pdb Write DModules (view mName m) mdata
+
+      evalWrite (_mInfo m) pdb Write DModuleSource (getHashedModuleName m) (ModuleCode code)
+
       -- Get all
       let newTopLevel = M.fromList $ (\(fqn, d) -> (_fqName fqn, (fqn, defKind (_mName m) d))) <$> fqDeps
           loadNewModule =
             over loModules (M.insert (_mName m) mdata) .
-            over loAllLoaded (M.union newLoaded) .
```
jmcardon, line 236: This line was removed by the way since on line 222, the following happens:
```
-- working set uses `lookupFqNameOrFail` and it _will_ lookup these deps
      esLoaded . loAllLoaded %= M.union newLoaded
```

So it is redundant.

rsoeldner (approved):

edmundnoble (approved):

PR #357, opened 2025-03-26
Fix syntax typo

lsgunnlsgunn:
PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/docs/builtins/General/try.md
+++ b/docs/builtins/General/try.md
@@ -4,11 +4,12 @@ Use the `try` function to attempt a **pure** action, returning a default value i
 Pure functions and pure expressions perform operations that produce a resulting value with no side effects. 
 
 Pure functions always return identical results for identical arguments.
-Pure expressions don't allow mutable variables reference arguments, or input and output operations.
+Pure expressions don't allow mutable variables, reference arguments, or input and output operations.
 
 Unlike impure expressions that support reading and writing to tables and working with non-deterministic state, pure expressions:
 
-- Don't write to memory or perform input or output operations.
+- Don't write to memory.
+- Don't perform input or output operations.
 - Don't work with non-deterministic state. 
 
 ### Basic syntax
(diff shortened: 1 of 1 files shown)


rsoeldner (approved):

docs/builtins/General/try.md
```diff
@@ -4,19 +4,20 @@ Use the `try` function to attempt a **pure** action, returning a default value i
 Pure functions and pure expressions perform operations that produce a resulting value with no side effects. 
 
 Pure functions always return identical results for identical arguments.
-Pure expressions don't allow mutable variables reference arguments, or input and output operations.
+Pure expressions don't allow mutable variables, reference arguments, or input and output operations.
 
 Unlike impure expressions that support reading and writing to tables and working with non-deterministic state, pure expressions:
 
-- Don't write to memory or perform input or output operations.
+- Don't write to memory.
```
jmcardon, line 11: don't write to the database**

CryptoPascal31:
Anyway... This help page (as most in Pact 5.0) is misleading contain many mistakes and looks like it was written by ChatpGPT.

https://github.com/kadena-io/pact-5/blob/641189099de0dd5920366c01e854f3008f7767d3/docs/builtins/General/try.md?plain=1#L6

This is not true, since (try) accepts Read-Only pure actions, and a value in a table may have changed between two calls of the same action.

https://github.com/kadena-io/pact-5/blob/641189099de0dd5920366c01e854f3008f7767d3/docs/builtins/General/try.md?plain=1#L7

- mutable variables => WTF in Pact ?

- reference arguments => What are references arguments in Pact ? The single type of references I know in Pact are modrefs and they are allowed in try statements.

- input oprations => AFAIK, input operations like (read-msg) are allowed in try statements.

jmcardon:
> This is not true, since (try) accepts Read-Only pure actions, and a value in a table may have changed between two calls of the same action.

@CryptoPascal31 you're actually correct. `Pure` from the POV of pact only concerns db reads.

PR #368, opened 2025-05-09
Fix recursion check in direct style interpreter

jmcardon:
Thanks @ggobugi27 for finding this bug

PR checklist:

* [ ] Benchmark regressions

### DIFF
--- a/pact-repl/Pact/Core/Repl.hs
+++ b/pact-repl/Pact/Core/Repl.hs
@@ -104,7 +104,7 @@ runRepl = do
       Just rest -> catch' $ do
         let src = accumulator <> "\n" <> rest
         lift (replCurrSource .== defaultSrc{_scPayload=src})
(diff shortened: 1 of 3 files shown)


pact-repl/Pact/Core/Repl.hs
```diff
@@ -104,7 +104,7 @@ runRepl = do
       Just rest -> catch' $ do
         let src = accumulator <> "\n" <> rest
         lift (replCurrSource .== defaultSrc{_scPayload=src})
-        eout <- lift (tryError (interpretReplProgramDirect (SourceCode "(interactive)" src)))
+        eout <- lift (tryError (interpretReplProgramBigStep (SourceCode "(interactive)" src)))
```
rsoeldner, line 107: I wonder why we change the interpreter?

rsoeldner (approved):

PR #369, opened 2025-05-09
Fix type of with-default read in typechecker, fix defun inference to be bidirectional

jmcardon:
Fixes w-d-r so subrows typecheck correctly, example
```pact
(module test-tc-with-default-read-subrow g
  (defcap g () true)
  (defschema foo a:integer b:integer c:string)
  (deftable footable:{foo})

  (defun subrow-typechecks-default-read ()
    (with-default-read footable "stuart"
      {"a":1, "c":"hello"}
      {"a" := a, "c" := c}
      (format "there are {} of {}" [a c])
    )
  ))

(typecheck "test-tc-with-default-read-subrow")
```

Also fixes return type bidirectionality in defuns so module references infer a bit better, so the following also typechecks
```pact
(interface foo-callable1
  (defun foo-callable1:bool (s:string)))


(interface foo-callable2
  (defun foo-callable2:bool (s:string)))

(module foo-callable-impl g
  (defcap g () true)
  (implements foo-callable1)
  (implements foo-callable2)

  (defun foo-callable1:bool (s:string)
    true)
  (defun foo-callable2:bool (s:string) true)

  (defun is-foo-callable:module{foo-callable1} ()
    foo-callable-impl)
  )

(typecheck "foo-callable-impl")
```

PR checklist:

* [x] Benchmark regressions
N/A
REPL only, doesn't matter

### DIFF
--- a/pact-repl/Pact/Core/Typed/Infer.hs
+++ b/pact-repl/Pact/Core/Typed/Infer.hs
@@ -753,10 +753,12 @@ instance TypeOfBuiltin CoreBuiltin where
       pure $ TypeScheme [r1, r2] [rowConstr] fnTy
     -- with-default-read : forall (r1: ROW) . table<r1> -> string -> object<r1> -> (object<r1> -> a) -> a
     CoreWithDefaultRead -> do
-      let r1 = RowVariable 1 "r1"
+      let r1 = RowVariable 2 "r1"
+          r2 = RowVariable 1 "r2"
       let aVar = TypeVariable 0 "a1"
-      let fnTy = TyTable (RowVar r1) :~> TyString :~> TyObject (RowVar r1) :~> (TyObject (RowVar r1) :~> TyVar aVar) :~> TyVar aVar
-      pure $ TypeScheme [r1, aVar] [] fnTy
+      let fnTy = TyTable (RowVar r1) :~> TyString :~> TyObject (RowVar r2) :~> (TyObject (RowVar r2) :~> TyVar aVar) :~> TyVar aVar
+          constr = RoseSubRow (RoseRowTy (RowVar r2)) (RoseRowTy (RowVar r1))
+      pure $ TypeScheme [r1, r2, aVar] [constr] fnTy
     CoreWithRead -> do
       let r1 = RowVariable 1 "r1"
       let aVar = TypeVariable 0 "a1"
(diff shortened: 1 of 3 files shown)


pact-repl/Pact/Core/Typed/Infer.hs
```diff
@@ -753,10 +753,12 @@ instance TypeOfBuiltin CoreBuiltin where
       pure $ TypeScheme [r1, r2] [rowConstr] fnTy
     -- with-default-read : forall (r1: ROW) . table<r1> -> string -> object<r1> -> (object<r1> -> a) -> a
     CoreWithDefaultRead -> do
-      let r1 = RowVariable 1 "r1"
+      let r1 = RowVariable 2 "r1"
+          r2 = RowVariable 1 "r2"
```
rsoeldner, line 757:
```suggestion
      let r1 = RowVariable 1 "r1"
          r2 = RowVariable 2 "r2"
```

pact-repl/Pact/Core/Typed/Infer.hs
```diff
@@ -753,10 +753,12 @@ instance TypeOfBuiltin CoreBuiltin where
...
       let aVar = TypeVariable 0 "a1"
-      let fnTy = TyTable (RowVar r1) :~> TyString :~> TyObject (RowVar r1) :~> (TyObject (RowVar r1) :~> TyVar aVar) :~> TyVar aVar
-      pure $ TypeScheme [r1, aVar] [] fnTy
+      let fnTy = TyTable (RowVar r1) :~> TyString :~> TyObject (RowVar r2) :~> (TyObject (RowVar r2) :~> TyVar aVar) :~> TyVar aVar
+          constr = RoseSubRow (RoseRowTy (RowVar r2)) (RoseRowTy (RowVar r1))
```
rsoeldner, line 760: Probably the other has to change too?
```suggestion
          constr = RoseSubRow (RoseRowTy (RowVar r1)) (RoseRowTy (RowVar r2))
```

rsoeldner (approved):

pact-repl/Pact/Core/Typed/Infer.hs
jmcardon: This is passing the test, so no

PR #370, opened 2025-05-13
fix repl autocomplete on special forms

jmcardon:
PR checklist:

* [x] Benchmark regressions

### DIFF
--- a/pact-repl/Pact/Core/Repl/Utils.hs
+++ b/pact-repl/Pact/Core/Repl/Utils.hs
@@ -208,9 +208,11 @@ replCompletion natives =
     prefixedNames <- uses (loaded . loModules) toPrefixed
     let
       allNames = Set.fromList $ T.unpack <$> concat
-        [tlns, moduleNames, prefixedNames, natives]
+        [tlns, moduleNames, prefixedNames, natives, specialFormNames]
     pure $ simpleCompletion <$> Set.toList (Set.filter (str `isPrefixOf`) allNames)
   where
+  specialFormNames =
+    ["and", "or", "if",  "enforce", "enforce-one", "with-capability", "try", "create-user-guard"]
   defNames = \case
     ModuleData md _ ->
       Term.defName <$> Term._mDefs md


CryptoPascal31:
I don't know... But maybe you should in the same time, include those special forms to the output of "pact -b"

CryptoPascal31:
> @CryptoPascal31 Sorry was away last week.
> 
> I could I guess, although they're not exactly builtins. I'll update the PR before I merge it then

From your perspective (because you know what happen behind the scene), they are not builtins.. But from a Pact dev perspective they are like builtins..

Let me relate my real story :smile: :
   - I wanted to do some tests with user guards.. but I didn't exactly remember what was the right function to call (create-user-guar, create-function-guard, ... ?)
   - I went into the REPL, I wrote ```(creat```, and pressed TAB, TAB, TAB, TAB...  :confounded: :=> Nothing interesting appeared => "WTF :rage:?"
   - So then I used ```pact -b |grep create``` then ```pact -b |grep guard``` .. => Nothing interesting appeared => "WTF :rage:?"
   - At some point I guessed whether I was tired or was becoming crazy, or whether you removed user guards !!
   - I went to bed :wink:

jmcardon:
> > @CryptoPascal31 Sorry was away last week.
> > I could I guess, although they're not exactly builtins. I'll update the PR before I merge it then
> 
> From your perspective (because you know what happen behind the scene), they are not builtins.. But from a Pact dev perspective they are like builtins..
> 
> Let me relate my real story 😄 :
> 
> * I wanted to do some tests with user guards.. but I didn't exactly remember what was the right function to call (create-user-guar, create-function-guard, ... ?)
> * I went into the REPL, I wrote `(creat`, and pressed TAB, TAB, TAB, TAB...  😖 :=> Nothing interesting appeared => "WTF 😡?"
> * So then I used `pact -b |grep create` then `pact -b |grep guard` .. => Nothing interesting appeared => "WTF 😡?"
> * At some point I guessed whether I was tired or was becoming crazy, or whether you removed user guards !!
> * I went to bed 😉

Hahaha fair enough, I'll update the PR

PR #362, opened 2025-04-18
[Experimental] Non-reentrancy special form

jmcardon:
This PR is a proof of concept of an experimental feature for pact 5, which is module level non-reentrancy.

## Syntax

For some syntactic term `k`, 
```pact
(non-reentrant k)
```

## Semantics
For any term `k` (in this example, `(m::foo 1)`)  written within a module `foo`'s definition (defcap, defun, defpact), e.g:
```pact
  (defun definitely-not-reentrant (m:module{foo-callable})
    (with-capability (SENSITIVE)
      ; NAME IS TENTATIVE
      (non-reentrant (m::foo 1))
      (called-internal)
      )
  )
```
`(non-reentrant k)` guarantees that `k` does not call back into `foo` during evaluation of `k`. 

PR checklist:

* [x] Test coverage for the proposed changes
* [x] PR description contains example output from repl interaction or a snippet from unit test output
* [x] (If Relevant) Documentation has been (manually) updated at https://docs.kadena.io/pact

Additionally, please justify why you should or should not do the following:

* [x] Benchmark regressions
* [x] Confirm replay/back compat (Ignore until core release)
* [x] (For Kadena engineers) Run integration-tests against a Chainweb built with this version of Pact (Ignore until core release)

PR #373, opened 2025-05-27
[Experimental] Add read-only reentrancy guarding for all modref calls

jmcardon:
Similar to #362 , but without a special form.

## What does this PR do exactly?

It turns all module reference calls back into the calling module into read-only. That is, any function call via module references that calls back into the calling module itself will call any function in the calling module in read-only mode (So no DB modifications are possible).

## Example:

See the test: Stub

An inline example:

For some interface
```pact
(interface some-interface
  (defun some-function:integer (param:integer))
```
and some caller:

```pact
(module some-module GOV
  (defcap GOV () (error "non-upgradable") ; Pact 5.3 feature on master, (error "foo") same as (enforce false "foo") but with different type
  ...
  (defun vulnerable-function (m:module{some-interface})
     (with-capability (SENSITIVE-CAP)
       (m::some-function 42))))
       
  (defun sensitive-internal-function ()
    (require-capability (SENSITIVE-CAP))
    (modifies-database-dangerously))
```

After this PR, the call above to `(m::some-function 42)` now cannot modify the database of `some-module`, no matter the value of `m`. So any attacker module that implements our interface `some-interface` can no longer attack `some-module`

## Why?

Access to data is not illegal, but most reentrant calls like this that are via modref calls are in a lot of cases malicious. This is a nerf to module references that will likely never be noticed by any regular modules on our chain.

## Why not?

This may be a bit heavy handed. This PR exists to assess impact first via an on-chain replay before this makes it into mainline pact

PR checklist:

* [x] Benchmark regressions

### DIFF
--- a/pact-repl/Pact/Core/IR/Eval/Direct/Evaluator.hs
+++ b/pact-repl/Pact/Core/IR/Eval/Direct/Evaluator.hs
@@ -125,7 +125,7 @@ eval
   -> EvalM e b i PactValue
 eval purity benv term = do
   ee <- viewEvalEnv id
-  let directEnv = envFromPurity purity (DirectEnv mempty (_eePactDb ee) benv (_eeDefPactStep ee) False)
+  let directEnv = envFromPurity purity (DirectEnv mempty (_eePactDb ee) benv (_eeDefPactStep ee) mempty False)
   evaluate directEnv term >>= \case
     VPactValue pv -> pure pv
     _ -> throwExecutionError (view termInfo term) (EvalError "Evaluation did not reduce to a value")
@@ -144,7 +144,7 @@ evalWithinCap info purity benv (CapToken qualName vs) term = do
   ee <- viewEvalEnv id
   (_, mh) <- getDefCapQN info qualName
   let ct = CapToken (qualNameToFqn qualName mh) vs
-  let cekEnv = envFromPurity purity (DirectEnv mempty (_eePactDb ee) benv (_eeDefPactStep ee) False)
+  let cekEnv = envFromPurity purity (DirectEnv mempty (_eePactDb ee) benv (_eeDefPactStep ee) mempty False)
   evalCap (view termInfo term) cekEnv ct PopCapInvoke NormalCapEval term >>= \case
     VPactValue pv -> pure pv
     _ ->
@@ -159,7 +159,7 @@ interpretGuard
   -> EvalM e b i PactValue
 interpretGuard info bEnv g = do
   ee <- viewEvalEnv id
-  let eEnv = DirectEnv mempty (_eePactDb ee) bEnv (_eeDefPactStep ee) False
+  let eEnv = DirectEnv mempty (_eePactDb ee) bEnv (_eeDefPactStep ee) mempty False
   PBool <$> enforceGuard info eEnv g
 
 evalResumePact
@@ -171,7 +171,7 @@ evalResumePact
 evalResumePact info bEnv mdpe = do
   ee <- viewEvalEnv id
   let pdb = _eePactDb ee
-  let env = DirectEnv mempty pdb bEnv (_eeDefPactStep ee) False
+  let env = DirectEnv mempty pdb bEnv (_eeDefPactStep ee) mempty False
   resumePact info env mdpe >>= \case
     VPactValue pv -> pure pv
     _ ->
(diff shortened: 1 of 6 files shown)


CryptoPascal31:
To be honest, I'm not very comfortable with this, because it restricts an already existent feature.

Using a callback may be a feature and may be desired: especially when referenced modules are trusted.

I strongly prefer the solutions of https://github.com/kadena-io/pact-5/pull/362 and https://github.com/kadena-io/pact-5/pull/367, letting the Pact dev to take its own responsibility by protecting its modrefs calls with (pure) or (non-reentrant) or both.

Anyway, even with all your efforts, by nature, modrefs are and will always be tricky. So instead of adding restriction that cannot be bypassed, IMO it's better to make available the right tools (ie pure and non-reentrant) to control precisely what happens.

My 2 cents

CryptoPascal31:
> I mean the problem with `pure` is that the _entire call_ is `pure`. Usually that's not what you want in those scenarios.

I just meant that pure is complementary… It protects from different threats.
 
> The `non-reentrant` special form is good, but the question is whether this reentrancy should be the default, or not. We could have this PR + a form called `(reentrant k)` that enables it. I'm actually of the opinion that the latter might be the better form. It's less of a footgun. I suspect that in most scenarios, users would prefer reentrancy to NOT be the default.

Looking at it that way, it makes sense to me ... But how do you see this `(reentrant)` form:
   - would it apply only to modrefs calls ? What happens if used in other contexts ?
   - would it only enable reentrant calls targeting the current module ?
   - would it taint further down the call stack ?
   

I'm just hoping, that `(reentrant)` won't become too complex to use and then creates new "hard to spot" issues.

IMHO, The approach with `(non-reentrant)` has the advantage of being very simple and clear, and can be simply see as a "lock".

> My 2 cents is: which should be the behavior that by default protects our users best? From this POV, this is I think our major goal at the team working on Pact.

I 100% agree.  I'm trying to be constructive to make Pact better for all.

jmcardon:
> Looking at it that way, it makes sense to me ... But how do you see this `(reentrant)` form:
> 
> * would it apply only to modrefs calls ? What happens if used in other contexts ?
> * would it only enable reentrant calls targeting the current module ?
> * would it taint further down the call stack ?
> 

- There's no change to how calls to module functions happen if you use them directly by name in this PR. If you call a function directly, you're responsible for vetting your dependencies, and because pact links exactly the version at the deploy time, you can't get attacked by a malicious upgrade unless you upgrade to the malicious upgrade. 
- It would only enable reentrant calls targeting the current module, and other modules that want to allow it must upgrade to use `(reentrant k)` as well, but again, only on module ref calls. 
- I mean the only "taint" that would happen is if you're intending to create a complex flow of module reference calls, but, as we have seen with one particular project which I'm sure you're well familiar with because of the vulnerabilities you've found, complex flows of circularly dependent module reference calls are _not_ the intended way to write pact, because with capabilities in scope, there's a lot of leak that can happen and security that can be bypassed. 

> I'm just hoping, that `(reentrant)` won't become too complex to use and then creates new "hard to spot" issues.
> 

Do you really not find that it is currently the opposite right now? Right now, the issue is that module references are _already_ too complex to use in the presence of scoped capabilities because the capability scope is inherited from the current context, and that's something we actually consider a feature. 

I'm quite frankly, far less worried about breaking a use case that we don't even know exists, than protecting against a known attack vector for products that _already_ live on our chain.

Lastly, the only complexity would be with getting some strange circular flow to work. By making reentrant calls by default read-only, it by default protects callers from malicious references attempting privilege escalation.

> IMHO, The approach with `(non-reentrant)` has the advantage of being very simple and clear, and can be simply seen as a "lock".
> 

I think it's less clear to me at this point why a language that doesn't allow recursion in the first place allows reentrancy on virtual calls, when this is one of our largest sources of vulnerabilities in the ecosystem.

CryptoPascal31:
> I think it's less clear to me at this point why a language that doesn't allow recursion in the first place allows reentrancy on virtual calls, when this is one of our largest sources of vulnerabilities in the ecosystem.

Agree.. I'm convinced..  :+1:

jwiegley (approved):
After a phone meeting on the aspects of this PR with @jmcardon, I give my approval to the approach and the reduction in functionality. Once CI is green and it's ready to merge in, you have my approval for delivery in the next service update.

PR #379, opened 2025-07-24
Add better example and xref to enforce-verifiers

lsgunnlsgunn:
Suggested improvements from Jermaine.

PR checklist:

* [ ] Benchmark regressions

---
  - https://app.asana.com/0/0/1210876821915459

### DIFF
--- a/docs/builtins/General/enforce-verifier.md
+++ b/docs/builtins/General/enforce-verifier.md
@@ -48,7 +48,7 @@ The following example demonstrates the use of a `signed_list` verifier plugin th
 In this example, the capability K uses the `enforce-verifier` function with the `signed_list` verifier. 
 The arguments that are passed to the capability are then evaluated by the verifier to ensure the integrity and authenticity of the message.
 
-The following example illustrates using the enforce-verifier in the capability definition, then using the `env-verifiers` function in the Pact REPL to add the "COOLZK" and "HYPERCHAIN-BRIDGE" verifier plugins to the environment data.
+The following example illustrates using the `enforce-verifier` function in the capability definition, then using the `env-verifiers` function in the Pact REPL to add the "COOLZK" and "HYPERCHAIN-BRIDGE" verifier plugins to the environment data.
 
 ```pact
 (module accounts GOV 
--- a/docs/builtins/General/enforce-verifier.md
+++ b/docs/builtins/General/enforce-verifier.md
@@ -2,6 +2,10 @@
 
 Use `enforce-verifier` to enforce that a verifier plugin with the specified `verifier-name` is in scope.
 
+Note that verifier plugins aren't implemented as Pact code.
+Instead, implementing a verifier plugin requires changes to the Chainweb protocol.
+For information about the design and implementation of verifier plugins, see [KIP-0028: Pact Verifier Plugins](https://github.com/kadena-io/KIPs/blob/master/kip-0028.md).
+
 ### Basic syntax
 
 To enforce that a verifier is in scope, use the following syntax:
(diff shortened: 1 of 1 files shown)


docs/builtins/General/enforce-verifier.md
```diff
@@ -24,14 +24,66 @@ The `enforce-verifier` function returns a boolean value indicating whether the s
...
+  (defcap K (msg:list dst-fin-id:string)
+    (enforce-verifier "signed_list")
+  )
+
+  (defun x ()
+    (with-capability 
+      (K [["issue","finp2p","citi:102:d0c3eb56-0fff-4670-adfd-ad291a4314c3",
+           "finId","02fd7923740a775c95ce17e9bb7239ff9096689f70db9263a7efb9a9ad08e9fed7","1"]]
+         "02fd7923740a775c95ce17e9bb7239ff9096689f70db9263a7efb9a9ad08e9fed7")
+      )
```
emilypi, line 43: The only critique i have here is that `K` and `x` are opaque names that maybe we should just name something because variables are hard for people to reason about sometimes.

docs/builtins/General/enforce-verifier.md
```diff
@@ -24,14 +24,66 @@ The `enforce-verifier` function returns a boolean value indicating whether the s
 
 ### Examples
```
rsoeldner, line 25: I feel like we should highlight that `verifier plugins` are implemented in chainweb. We currently *only* provide `hyperlane_v3_message` (see https://github.com/kadena-io/chainweb-node/blob/e0acda06fcda6217e2241b728f266199b39bee42/src/Chainweb/Version/Mainnet.hs#L222C55-L222C75)

The example sections always assume a particular verifier to be implemented, in this new case the `signed_list`.
Within the pact repl, all verifiers can be simulated using the `env-verifiers` builtin - but they are not available on-chain (mainnet, devnet, ...).

Maybe @edmundnoble can comment here. My general understanding is that verifier plugins are a light-weight addition we can make to (mainly) partners to provide more complex integrations\infrastructure.

docs/builtins/General/enforce-verifier.md
```diff
@@ -24,14 +24,66 @@ The `enforce-verifier` function returns a boolean value indicating whether the s
...
+      (K [["issue","finp2p","citi:102:d0c3eb56-0fff-4670-adfd-ad291a4314c3",
+           "finId","02fd7923740a775c95ce17e9bb7239ff9096689f70db9263a7efb9a9ad08e9fed7","1"]]
+         "02fd7923740a775c95ce17e9bb7239ff9096689f70db9263a7efb9a9ad08e9fed7")
+      )
+      1)
+)
+```
+
+In this example, the capability K uses the `enforce-verifier` function with the `signed_list` verifier. 
+The arguments that are passed to the capability are then evaluated by the verifier to ensure the integrity and authenticity of the message.
```
rsoeldner, line 49: The verifier plugin can fail the transaction, which will result in the cap. not being granted.

docs/builtins/General/enforce-verifier.md
lsgunnlsgunn: Added a pointer to the KIP

docs/builtins/General/enforce-verifier.md
```diff
@@ -2,9 +2,13 @@
 
 Use `enforce-verifier` to enforce that a verifier plugin with the specified `verifier-name` is in scope.
 
+Note that verifier plugins aren't implemented as Pact code.
+Instead, implementing a verifier plugin requires changes to the Chainweb protocol.
```
rsoeldner, line 6: this is great, thank you

rsoeldner (approved):

